PR #9, opened 2015-07-01
OKHTTP upgrade

kingofirony:
- Upgrades okhttp to version 2.4


ecgreb:
No longer needed. OkHttp is included as a transitive dependency only.

PR #10, opened 2015-07-02
adds footer and fixes view transitions

kingofirony:
- Fixes transition between route preview and route mode
- Adds route footer in route mode


app/src/main/res/layout/view_route_mode.xml
```diff
@@ -1,11 +1,56 @@
...
-        android:layout_height="75dp" />
+        android:layout_height="@dimen/pager_height"/>
+
+    <RelativeLayout
+        android:layout_width="match_parent"
+        android:layout_height="match_parent"
+        android:layout_gravity="bottom"
+        android:layout_alignParentBottom="true">
+
+        <LinearLayout
```
ecgreb: Do we need a LinearLayout inside a RelativeLayout here? Can these be collapsed into a single container for the footer?

PR #14, opened 2015-07-10
Map location updates in routing mode

ecgreb:


app/src/main/kotlin/com/mapzen/erasermap/view/MainActivity.kt
```diff
@@ -45,6 +45,11 @@ import javax.inject.Inject
 
 public class MainActivity : AppCompatActivity(), ViewController, Router.Callback,
         SearchResultsView.OnSearchResultSelectedListener {
+    companion object {
+        val DEFAULT_ZOOM: Float = 14f
+        val ROUTING_ZOOM: Float = 17f
+    }
+
```
kingofirony: 👍

app/src/main/kotlin/com/mapzen/erasermap/view/MainActivity.kt
```diff
@@ -146,18 +151,32 @@ public class MainActivity : AppCompatActivity(), ViewController, Router.Callback
         LocationServices.FusedLocationApi?.requestLocationUpdates(locationRequest) {
             location: Location -> currentLocation = location
             val routeModeView = findViewById(R.id.route_mode) as RouteModeView
-            routeModeView.onLocationChanged(location)
+            if (routeModeView != null && routeModeView.getVisibility() == View.VISIBLE) {
```
kingofirony: maybe could have a check in the presenter since it keeps track of routing mode being enabled?

ecgreb: Good suggestion. Using view state to control the flow of data in the app is gross.

I added a commit that moves location update logic into the presenter (with tests)

PR #24, opened 2015-07-24
Peter/key get

kingofirony:
- Gets keys from server

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -1,8 +1,6 @@
 <?xml version="1.0" encoding="utf-8"?>
 <manifest xmlns:android="http://schemas.android.com/apk/res/android"
-    package="com.mapzen.erasermap"
-    android:versionCode="2"
-    android:versionName="0.1">
+    package="com.mapzen.erasermap">
 
     <uses-feature android:glEsVersion="0x00020000" android:required="true" />
 
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -1,6 +1,8 @@
 <?xml version="1.0" encoding="utf-8"?>
 <manifest xmlns:android="http://schemas.android.com/apk/res/android"
-    package="com.mapzen.erasermap">
+    package="com.mapzen.erasermap"
+    android:versionCode="2"
+    android:versionName="0.1">
 
     <uses-feature android:glEsVersion="0x00020000" android:required="true" />
 
--- a/app/src/main/java/com/mapzen/erasermap/CommonModule.java
+++ b/app/src/main/java/com/mapzen/erasermap/CommonModule.java
@@ -27,9 +27,7 @@ public class CommonModule {
         return new Bus();
     }
 
-    @Provides @Singleton ManifestModel provideApiKeys() { return new ManifestModel(); }
+    @Provides @Singleton ManifestModel provideManifestModel() { return new ManifestModel(); }
 
-    @Provides RouteEngine provideRouteEnCgine() {
-        return new RouteEngine();
-    }
+    @Provides RouteEngine provideRouteEngine() { return new RouteEngine(); }
 }
--- a/app/src/main/java/com/mapzen/erasermap/CommonModule.java
+++ b/app/src/main/java/com/mapzen/erasermap/CommonModule.java
@@ -1,5 +1,6 @@
 package com.mapzen.erasermap;
 
+import com.mapzen.erasermap.model.ManifestModel;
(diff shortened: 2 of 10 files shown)


app/src/main/AndroidManifest.xml
```diff
@@ -1,6 +1,8 @@
 <?xml version="1.0" encoding="utf-8"?>
 <manifest xmlns:android="http://schemas.android.com/apk/res/android"
-    package="com.mapzen.erasermap">
+    package="com.mapzen.erasermap"
+    android:versionCode="2"
+    android:versionName="0.1">
```
ecgreb: No need to declare version code and version name here. It gets injected by Gradle at build time.

app/src/main/java/com/mapzen/erasermap/CommonModule.java
```diff
@@ -26,7 +27,9 @@
         return new Bus();
     }
 
-    @Provides RouteEngine provideRouteEngine() {
+    @Provides @Singleton ManifestModel provideApiKeys() { return new ManifestModel(); }
```
ecgreb: Typically Dagger provider methods will mirror name of return type. For example `provideManifestModel()`.

app/src/main/java/com/mapzen/erasermap/CommonModule.java
```diff
@@ -26,7 +27,9 @@
...
+
+    @Provides RouteEngine provideRouteEnCgine() {
```
ecgreb: Typo

app/src/main/kotlin/com/mapzen/erasermap/view/MainActivity.kt
```diff
@@ -90,6 +93,7 @@ public class MainActivity : AppCompatActivity(), MainViewController, Router.Call
         initReverseButton()
         centerMapOnCurrentLocation()
         presenter?.onRestoreViewState()
+        getApiKeys({initCrashReportService()})
```
ecgreb: Does this mean initializing crash reports doesn't happen until the API manifest is downloaded? What if the app crashes before the download completes? Or the download itself crashes the app? I think we would still want to be able to catch those errors.

Perhaps we want to continue to inject the MINT_API_KEY via BuildConfig and reserve the API manifest for our services (Vector Tiles, Pelias, Valhalla).

app/src/main/kotlin/com/mapzen/erasermap/view/MainActivity.kt
```diff
@@ -65,6 +67,8 @@ public class MainActivity : AppCompatActivity(), MainViewController, Router.Call
       @Inject set
     var crashReportService: CrashReportService? = null
       @Inject set
+    var apiKeys: ManifestModel? = null
+      @Inject set
```
ecgreb: Do we need to inject an empty `ManifestModel` here when a new instance is created by `ManifestDownloader`?

app/src/test/java/com/mapzen/erasermap/model/ManifestDownLoaderTest.kt
```diff
@@ -0,0 +1,72 @@
...
+
+    }
+
+    After
+    throws(Exception::class)
+    public fun teardown() {
+        server?.shutdown()
+    }
+
+    Test
```
ecgreb: Wow the `@` symbol is optional for JUnit annotations... weird. Nice job getting tests to run in Kotlin.

kingofirony: Done

app/src/main/java/com/mapzen/erasermap/CommonModule.java
kingofirony: Done

app/src/main/java/com/mapzen/erasermap/CommonModule.java
kingofirony: Done

app/src/main/AndroidManifest.xml
kingofirony: Done

app/src/main/kotlin/com/mapzen/erasermap/view/MainActivity.kt
kingofirony: Done

app/src/main/kotlin/com/mapzen/erasermap/view/MainActivity.kt
kingofirony: Done

PR #25, opened 2015-07-27
Continue routing while app is in background

ecgreb:
* Extracts MapzenLocation model object
* Maintains location updates in background while routing

### DIFF
--- a/app/src/main/kotlin/com/mapzen/erasermap/model/MapzenLocation.kt
+++ b/app/src/main/kotlin/com/mapzen/erasermap/model/MapzenLocation.kt
@@ -1,0 +1,57 @@
+package com.mapzen.erasermap.model
+
+import android.location.Location
+import com.mapzen.android.lost.api.LocationRequest
+import com.mapzen.android.lost.api.LocationServices
+import com.mapzen.android.lost.api.LostApiClient
+import com.mapzen.erasermap.EraserMapApplication
+import com.mapzen.pelias.PeliasLocationProvider
+import javax.inject.Inject
+
+public class MapzenLocation (val app: EraserMapApplication) : PeliasLocationProvider {
+    private val LOCATION_UPDATE_INTERVAL_IN_MS: Long = 1000L
+    private val LOCATION_UPDATE_SMALLEST_DISPLACEMENT: Float = 0f
+
+    var locationClient: LostApiClient? = null
+        @Inject set
+
+    init {
+        app.component()?.inject(this)
+    }
+
+    public fun connect() {
+        locationClient?.connect()
+    }
+
+    public fun disconnect() {
+        locationClient?.disconnect()
+    }
+
+    public fun initLocationUpdates(callback: (location: Location) -> Unit) {
+        if (locationClient?.isConnected() == false) {
+            locationClient?.connect()
+        }
+
+        val locationRequest = LocationRequest.create()
+                .setPriority(LocationRequest.PRIORITY_HIGH_ACCURACY)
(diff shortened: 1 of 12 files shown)


app/src/main/kotlin/com/mapzen/erasermap/model/MapzenLocation.kt
```diff
@@ -0,0 +1,12 @@
...
+import android.location.Location
+import com.mapzen.pelias.PeliasLocationProvider
+
+public interface MapzenLocation : PeliasLocationProvider {
+    public fun connect()
+    public fun disconnect()
+    public fun isConnected(): Boolean
+    public fun initLocationUpdates(callback: (location: Location) -> Unit)
+    public fun getLastLocation(): Location?
+}
```
kingofirony: Using the model more - yesssss.

app/src/main/kotlin/com/mapzen/erasermap/model/MapzenLocationImpl.kt
```diff
@@ -0,0 +1,62 @@
...
+        locationClient?.disconnect()
+    }
+
+    override fun isConnected(): Boolean {
+        val isConnected = locationClient?.isConnected()
+        if (isConnected is Boolean) return isConnected else return false
+    }
+
+    override fun initLocationUpdates(callback: (location: Location) -> Unit) {
+        if (locationClient?.isConnected() == false) {
```
kingofirony: !locationClient?.isConnected() is the same thing no?

kingofirony: picky I know no real need to change it.

app/src/main/kotlin/com/mapzen/erasermap/presenter/MainPresenterImpl.kt
```diff
@@ -166,4 +168,20 @@ public class MainPresenterImpl() : MainPresenter {
         mainViewController?.setMapTilt(MainPresenter.ROUTING_TILT)
         mainViewController?.setMapRotation(Math.toRadians(instruction.bearing.toDouble()).toFloat())
     }
+
+    override fun onPause(mapzenLocation: MapzenLocation?) {
+        if (isRouting() || isRoutingDirectionList()) {
+            return
+        }
+
+        mapzenLocation?.disconnect()
```
kingofirony: if not routing AND not routing direction list would do the same thing in 3 lines

app/src/main/kotlin/com/mapzen/erasermap/model/MapzenLocationImpl.kt
ecgreb: Good point.

Unfortunately you can't put a `!` in front of `locationClient?.isConnected()` because it returns type `Boolean?`

However just calling `isConnected()` on `this` does work since it returns type `Boolean` (not nullable). I will push an update.

kingofirony: there's .not() function if you wanted to use it

app/src/main/kotlin/com/mapzen/erasermap/presenter/MainPresenterImpl.kt
ecgreb: Done

app/src/main/kotlin/com/mapzen/erasermap/model/MapzenLocationImpl.kt
kingofirony: ah trying it out now turns out you can't on boolean?. bummer

PR #37, opened 2015-09-15
Maintain route state on rotation

ecgreb:
Prevents replay of completed instructions when rotating the device during routing.

* Make RouteEngine a singleton
* Extracts RoutePresenter
* Do not reset RouteEngine if already routing
* Verify icon is ImageView to prevent type cast error


ecgreb:
cc @dgearhart

PR #39, opened 2015-09-15
Scene file and poi texture

tallytalwar:
- This should be updated as tangram-es library updates and supports more style sheet functionality


ecgreb:
Thanks @tallytalwar !

tallytalwar:
:metal:

PR #50, opened 2015-09-29
fixes bug involving lack of gradle.properties

kingofirony:
- Fixes bug involving crash if buildconfig is not available.


ecgreb:
Looks like this one was on me. Nice catch :)

ecgreb:
cc @tallytalwar @blair1618

tallytalwar:
will give this a shot before the meeting today.

tallytalwar:
This works ... Thanks for the fix :D.

PR #53, opened 2015-10-01
Appends vector tile API key to tile requests using HttpHandler

ecgreb:


app/src/main/kotlin/com/mapzen/erasermap/view/MainActivity.kt
```diff
@@ -73,6 +74,8 @@ public class MainActivity : AppCompatActivity(), MainViewController, RouteCallba
         @Inject set
     var settings: AppSettings? = null
         @Inject set
+    var tileHttpHandler: TileHttpHandler? = null
+        @Inject set
```
tallytalwar: is this some Kotlin thingy?

tallytalwar:
Looks good to me.

app/src/main/kotlin/com/mapzen/erasermap/view/MainActivity.kt
ecgreb: `@Inject set` tells Dagger to inject this dependency via its Kotlin setter.

blair1618:
This looks fine to me - can we merge it so that the automated builds start showing tiles again? :)

I will also add a default HTTP handler like you mentioned yesterday, that won't require any changes on the eraser-map side.

PR #56, opened 2015-10-05
Peter/pelias v1

kingofirony:
- Implements new pelias-android-sdk that uses pelias-v1

### DIFF
--- a/app/src/main/kotlin/com/mapzen/erasermap/view/MainActivity.kt
+++ b/app/src/main/kotlin/com/mapzen/erasermap/view/MainActivity.kt
@@ -416,11 +416,13 @@ public class MainActivity : AppCompatActivity(), MainViewController, RouteCallba
 
     override fun centerOnCurrentFeature(features: List<Feature>) {
         Handler().postDelayed({
-            val pager = findViewById(R.id.search_results) as SearchResultsView
-            val position = pager.getCurrentItem()
-            val feature = SimpleFeature.fromFeature(features.get(position))
-            mapController?.setMapPosition(feature.lon,feature.lat)
-            mapController?.mapZoom = MainPresenter.DEFAULT_ZOOM
+            if(features.size() > 0) {
+                val pager = findViewById(R.id.search_results) as SearchResultsView
+                val position = pager.getCurrentItem()
+                val feature = SimpleFeature.fromFeature(features.get(position))
+                mapController?.setMapPosition(feature.lon, feature.lat)
+                mapController?.mapZoom = MainPresenter.DEFAULT_ZOOM
+            }
         }, 100)
     }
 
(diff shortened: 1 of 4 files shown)


app/src/main/kotlin/com/mapzen/erasermap/view/MainActivity.kt
```diff
@@ -11,10 +11,7 @@ import android.preference.PreferenceManager
 import android.support.v4.view.MenuItemCompat
 import android.support.v7.app.AppCompatActivity
 import android.util.Log
-import android.view.Menu
-import android.view.MenuItem
-import android.view.MotionEvent
-import android.view.View
+import android.view.*
```
ecgreb: Let's avoid using * imports since they can be ambiguous. I think IntelliJ/Studio uses * imports by default if there are 5 or more classes from the same package but you can change this in the IDE settings.

app/src/main/kotlin/com/mapzen/erasermap/view/MainActivity.kt
```diff
@@ -700,12 +701,22 @@ public class MainActivity : AppCompatActivity(), MainViewController, RouteCallba
         routeModeView.route = null
     }
 
+    private fun setBoundingBox() {
+        val menuItem = optionsMenu?.findItem(R.id.action_search)
+        var mdisp : Display = getWindowManager().getDefaultDisplay();
+        var minLatLon = mapController?.coordinatesAtScreenPosition(mdisp.width.toDouble(), mdisp.height.toDouble())
+        var maxLatLon = mapController?.coordinatesAtScreenPosition(0.0, 0.0);
+        var bbox: BoundingBox = BoundingBox(minLatLon?.latitude as Double, minLatLon?.longitude as Double,
+                maxLatLon?.latitude as Double, maxLatLon?.longitude as Double)
```
blair1618: Probably makes sense to take the `min` and `max` of the two latitudes and longitudes when creating the bounding box, either here or within the `BoundingBox` constructor. If the map view is rotated, the left side of the screen may no longer be the West-most side.

ecgreb: I just had a conversation with @riordan and @blair1618 and there are a few potential issues with how we are currently using the bounding box in search.

1. We should not be using the `boundary.rect` query param. This will perform a strict search and only return results inside the boundary. Instead we should be using `focus.viewport` which will inform the search but not strictly limit it. This needs to be updated in [`PeliasService`](https://github.com/mapzen/pelias-android-sdk/blob/master/src/main/java/com/mapzen/pelias/PeliasService.java).

2. Pelias is assuming bottom-left and top-right coordinates for the rect. Given that Android screen coordinates originate at the top-left corner of the display we should not be passing the coordinates at `(0,0)` and `(display.width, display.height)` but instead `(0, display.height)` and `(display.width, 0)`.

3. Finally since the user can rotate the map we cannot assume that the bottom left corner is the minimum lat/lon geographically. What if the user rotates the map 180 degrees so South is at the top of the screen? As stated by @blair1618 above we should compare the lat/lon values obtained at each corner and set the lower values as the min and higher values as the max. I like the idea of adding this logic to the `BoundingBox` constructor.

PR #57, opened 2015-10-06
Allow user to interact with map while routing

ecgreb:
* Adds logic to show/hide resume button while routing
* Allows user to interact with map in routing mode
* Publishes location updates on the event bus
* Revises `MapzenLocation` API and implementation
* Tracks user location while routing


ecgreb:
Requires https://github.com/tangrams/tangram-es/pull/328

PR #60, opened 2015-10-12
Snap location and rotation

ecgreb:
* Use snap location to set icon position in routing mode.
* Add separate MapData layer for routing icon.
* Hide find me icon while in routing mode.
* Hide route icon on exit routing mode.


ecgreb:
Requires https://github.com/mapzen/on-the-road/pull/46

riordan:
This should make a massive difference for the driving experience.

PR #63, opened 2015-10-13
Update the scene file to make routing icon

tallytalwar:
transition of hide->show 0sec

- Still better than previous, more work required to make sure old route marker location is not cleared before the new
one gets fetched and added.


ecgreb:
Nice! Can you add the same treatment to the `find_me` icon?

tallytalwar:
Sure.

tallytalwar:
I think search icons should also get this?

PR #70, opened 2015-10-26
Replace fixed style assets with submodule

blair1618:
This pulls in the `android` branch of the eraser-map style repository as a submodule


ecgreb:
:+1:

PR #79, opened 2015-10-29
Sign all debug builds with same keystore

ecgreb:
Prevents having to uninstall the previous version before installing latest APK. Uses debug.keystore file automatically generated by Circle CI.


tallytalwar:
This will be very helpful for testing :D.

Issue #103, opened 2015-11-04
Unable to install on some JellyBean and ICS devices


ecgreb:
If initial app install fails on some JellyBean devices (insufficient storage, dex opt error) the app can never be re-installed due to files left behind.

Fix requires factory reset or root access to delete leftover files.

For more information see https://code.google.com/p/android/issues/detail?id=65790

**Reproduced on**

Motorola Droid Razr
Android 4.1.2

Samsung Galaxy S3
Android 4.0.4

**Initial error**

```
Waiting for device.
Target device: motorola-droid_razr-015499070E007006
Uploading file
    local path: /Users/Varun/Development/eraser-map/app/build/outputs/apk/app-debug.apk
    remote path: /data/local/tmp/com.mapzen.erasermap
Installing com.mapzen.erasermap
DEVICE SHELL COMMAND: pm install -r "/data/local/tmp/com.mapzen.erasermap"
    pkg: /data/local/tmp/com.mapzen.erasermap
Failure [INSTALL_FAILED_DEXOPT]


DEVICE SHELL COMMAND: pm uninstall com.mapzen.erasermap
Unknown failure
```

**Subsequent errors**

```
$ adb install app/build/outputs/apk/app-debug.apk
4955 KB/s (9815087 bytes in 1.934s)
	pkg: /data/local/tmp/app-debug.apk
Failure [INSTALL_FAILED_UID_CHANGED]
rm failed for -f, No such file or directory
```

blair1618:
Just to chime in: I can also reproduce this with a dexopt error on a Galaxy Nexus running 4.2.2 
(I have root access to fix it though!)

This is due to the size of the S3 SDK, right? Do we have a viable alternative for getting keys?

ecgreb:
Working on a fix now. Removing AWS SDK and multidex. Settings API keys via `gradle.properties` and `BuildConfig.java` for now until we come up with a better solution.

PR #106, opened 2015-11-05
Fix install and launch issues on OS versions < KitKat

ecgreb:
* Temporarily remove leyndo-android-library from project
* Disable multidex support
* Set API keys via `gradle.properties` and `BuildConfig.java`
* Make kotlin-reflect test dependency only to reduce method count


ecgreb:
Closes https://github.com/mapzen/eraser-map/issues/103

Tracked down root cause to a bad version of Kotlin that was introducing dex errors.

https://devnet.jetbrains.com/message/5561799#5561799

PR #124, opened 2015-11-16
Updates the client datasource api usage

tallytalwar:
- Fixes route icon flickering
- NOTE: Currently lacks smooth transition between 2 route icon positions. Marking it as a TODO for further improvements.

Dependency: https://github.com/tangrams/tangram-es/pull/400


tallytalwar:
Ohh this is dependent on a pr on tangram which is still not merged. Can you
revert this till tangram master gets updated with these changes.

PR #125, opened 2015-11-17
Revert "Updates the client datasource api usage"

ecgreb:
Reverts mapzen/eraser-map#124


blair1618:
This revert needs to be merged to prevent crashes that are happening when displaying search results with the current build.

ecgreb:
Sorry for jumping the gun on this one first time around

tallytalwar:
Real cause of the crash is https://github.com/tangrams/tangram-es/issues/404
.

PR #128, opened 2015-11-20
Update style submodule; now using gh-pages branch

blair1618:
No more separate android branch!


tallytalwar:
Yay! Looks good to me. Will merge.

Issue #78, opened 2015-10-29
Add ability to reorient the map to North without changing the extent


rmglennon:
When zooming with multitouch, it's really easy (maybe too easy) to rotate the map. If you tap the compass rose in the bottom corner, the map reorients, but it also zooms back to the default home location and extent. It would be really handy to have a way to keep your current extent, but remove the rotation that is often added accidentally.

See snips from this chat:
Mike
we should also make it less easy to rotate the map when zooming
Rhonda
yep. there's a related issue for that:  Issue #68: Don't allow map rotation when zoomed (0,1,2,3,4,5) by nvkelso in mapzen/eraser-map on GitHub
Ekta
yeah Nathaniel and I made that issue last week
I talked to Varun about it and he said he will get to it
but adding a compass at all times is a VERY NICE TO HAVE
I have the design and asset ready… just have to build it ;)
Matt
We have all the capabilities we'd need for this; you can query or set the rotation of the tangram map view and add a button-like compass (like the "find me" button)
It will also help when we add smooth animations for view transitions (which will happen!)

bcamper:
Should be handled by the app until we build view animation into Tangram
(which we plan to do).

ecgreb:
This would be easy to do in the app.

@souperneon can you design an icon and placement for a find North button?

souperneon:
@ecgreb will do!

PR #163, opened 2015-12-01
new compass north

souperneon:
https://github.com/mapzen/eraser-map/issues/78 @ecgreb


souperneon:
I picked an icon that can be rotated around the center to indicate changing north as the user rotates the map and in it's default position points North

ecgreb:
@souperneon I think we might need to go a little bigger with the icon. Its kind of hard to click right now. Also the grey circle is difficult to see over the base map color.

image

souperneon:
@ecgreb I was thinking we add the same white square behind it? If that gets too clunky I can have the circle filled in? How about we try the square first?

ecgreb:
@souperneon sure we can try that. We'll also need pressed state assets for the icon.

Issue #174, opened 2015-12-07
Route preview view list shows SimpleFeature destination


nvkelso:
* **Device name:** Samsung Galaxy S6 edge
* **Android Version:** 5.1.1
* **App build number:** master-518
* **What did you expected to happen?** I should see my destination in the to/from box above the route maneuver list.
* **What happened instead?** I saw SimpleFeature{id=123} in the destination field instead of my destination street address.
* **Steps to reproduce:** Simulate location to San Francisco. Search for McKinleyville, Humboldt County, California. Route to the search result to preview the route. Tap the View List button.
* **Attach a screenshot** See below.
* **Attach device logs** eh.

nvkelso:
![screenshot_2015-12-06-22-33-26](https://cloud.githubusercontent.com/assets/853051/11620684/056f505c-9c6a-11e5-8bbe-862a207866ab.png)
![screenshot_2015-12-06-22-33-24](https://cloud.githubusercontent.com/assets/853051/11620685/0572d9b6-9c6a-11e5-89d9-0b27980ed18b.png)

ecgreb:
Fixed in #177

PR #177, opened 2015-12-07
Add compass view

### LINKED ISSUE #78: Add ability to reorient the map to North without changing the extent

ecgreb:
Adds a dynamic compass to all screens that allows the user to view current map orientation and reset to North.

* Show compass when map is rotated
* Hide compass when view is reset to North
* Reset map orientation and animate compass on click
* Show compass in default, search, route preview, and navigation modes
* Replaces feature `toString()` with `name()` in direction list header (#174)

Rather than attempting to re-position a single compass across multiple views both `view_route_preview.xml` and `view_route_mode.xml` declare their own separate instances of `CompassView` that are shown/hidden along with their parent views so only one is visible at a time. The combination of `MainActivity` and `MainPresenter` handle updating all 3 instances on map rotation and click.

Closes #78 and #174 

image


app/src/main/kotlin/com/mapzen/erasermap/view/InstructionListActivity.kt
```diff
@@ -31,13 +39,15 @@ public class InstructionListActivity : AppCompatActivity() {
         }
     }
 
-    private fun setHeaderOrigins(bundle: Bundle?, reverse: Boolean?) {
-        if (reverse == true) {
-            (findViewById(R.id.starting_point) as TextView).setText(bundle?.getString("destination"))
-            (findViewById(R.id.destination) as TextView).setText(R.string.current_location)
+    private fun setHeaderOrigins(destination: String, reverse: Boolean) {
+        val startTextView = findViewById(R.id.starting_point) as TextView
```
msmollin: Yay programming

app/src/main/kotlin/com/mapzen/erasermap/view/MainActivity.kt
```diff
@@ -77,15 +77,18 @@ public class MainActivity : AppCompatActivity(), MainViewController, RouteCallba
...
+    val routePreviewView: RoutePreviewView by lazy { findViewById(R.id.route_preview) as RoutePreviewView }
+    val routeModeView: RouteModeView by lazy { findViewById(R.id.route_mode) as RouteModeView }
+    val reverseButton: ImageButton by lazy { findViewById(R.id.route_reverse) as ImageButton }
+    val viewListButton: Button by lazy { findViewById(R.id.view_list) as Button }
+    val startNavigationButton: Button by lazy { findViewById(R.id.start_navigation) as Button }
+    val byCar: RadioButton by lazy { findViewById(R.id.by_car) as RadioButton }
+    val byBike: RadioButton by lazy { findViewById(R.id.by_bike) as RadioButton }
+    val byFoot: RadioButton by lazy { findViewById(R.id.by_foot) as RadioButton }
+    val compass: CompassView by lazy { findViewById(R.id.compass_view) as CompassView }
+    val routePreviewCompass: CompassView by lazy { findViewById(R.id.route_preview_compass_view) as CompassView }
```
msmollin: I know you probably already thought of this, but just incase you have the lazy instantiation can cause performance issues if the system is under stress for whatever reason and it goes to instantiate a performance / memory intensive component. I don't know if any of these are like that, but something to keep in mind.

app/src/main/kotlin/com/mapzen/erasermap/view/MainActivity.kt
```diff
@@ -164,8 +170,20 @@ public class MainActivity : AppCompatActivity(), MainViewController, RouteCallba
     private fun initFindMeButton() {
         findMe = MapData("find_me")
         Tangram.addDataSource(findMe);
-        findMeButton?.visibility = View.VISIBLE
-        findMeButton?.setOnClickListener({ presenter?.onFindMeButtonClick() })
+        findMeButton.visibility = View.VISIBLE
+        findMeButton.setOnClickListener({ presenter?.onFindMeButtonClick() })
```
msmollin: I guess I'm unsure why in some places here you've changed from explicitly calling the setter functions for properties, and in others you haven't. Is this on purpose?

app/src/main/kotlin/com/mapzen/erasermap/view/MainActivity.kt
```diff
@@ -455,11 +476,11 @@ public class MainActivity : AppCompatActivity(), MainViewController, RouteCallba
 
     override fun success(route: Route) {
         routeManager?.route = route
-        routePreviewView?.route = route
```
msmollin: I like the cleanup removing the optional checking here. However, ignorant Android question - findViewById will always return the correct type or crash, yes?

app/src/main/kotlin/com/mapzen/erasermap/view/InstructionListActivity.kt
ecgreb: \o/

app/src/main/kotlin/com/mapzen/erasermap/view/MainActivity.kt
ecgreb: Good point. In this case its not actually constructing the view lazily but rather doing a lookup to find it in the view hierarchy. The actual instance is created when the xml layout gets inflated.

The nice advantage of doing it this way is it delays the lookup until after the view is inflated so these properties don't have to be declared nullable.

app/src/main/kotlin/com/mapzen/erasermap/view/MainActivity.kt
ecgreb: Property access syntax when calling from Kotlin into Java was added after we started the app so a lot of those changes are just updating old code.

In this case Kotlin is unable to use property access syntax because `View.setOnClickListener(...)` is not a true setter - it does not actually set a field on itself but rather invokes 'getListenerInfo().mOnClickListener`.

https://android.googlesource.com/platform/frameworks/base/+/refs/heads/master/core/java/android/view/View.java#5126

app/src/main/kotlin/com/mapzen/erasermap/view/MainActivity.kt
ecgreb: Indeed. If the `findViewById(...)` lookup fails to find the expected type (in the lazy initializer above) the app will crash. It also means something is seriously wrong so crashing is probably the best option available at that point.

PR #185, opened 2015-12-08
Redraw route line on reroute

### LINKED ISSUE #97: Route highlight did not update on re-route
* **Android Version:** 5.0
* **App build number:** master-420
* **What did you expected to happen?** That when taking different path that the route will recalculate and display properly on the map
* **What happened instead?** The route was recalculated however the route highlight on the map was never updated. Also, sometimes there were several current location pins displayed on the map. Please see screenshots
* **Steps to reproduce:** Run a route, take a different path and notice the route highlight and current position icon is not correct on the map
* **Attach a screenshot**
screenshot_2015-11-03-16-24-16
screenshot_2015-11-03-16-26-35
screenshot_2015-11-03-16-57-16
screenshot_2015-11-03-16-58-31

ecgreb:
Closes #97


rmglennon:
Ah, this is a known issue.

https://github.com/mapzen/eraser-map/issues/187

PR #194, opened 2015-12-11
Refinements for interacting with map in routing mode

### LINKED ISSUE #74: Routing Mode: Snapping when panning
Mode: Routing
Situation: Panning around, while a turn or a routing event happens.
Action/Issue: Snaps to the current location on the route, instead of keeping the view at the panned location.

Phone: Samsung Galaxy S5
Android Version: 5.0
Date Tested: 10/28/2015

ecgreb:
* Track user swipe on view pager vs. programatic changes
* Do not show resume button unless map is moved minimum distance by user
* Do not reset map position to current instruction while user is panning the map

Closes #74

### DIFF
--- a/app/src/main/kotlin/com/mapzen/erasermap/presenter/RoutePresenter.kt
+++ b/app/src/main/kotlin/com/mapzen/erasermap/presenter/RoutePresenter.kt
@@ -6,13 +6,17 @@ import com.mapzen.valhalla.Instruction
 import com.mapzen.valhalla.Route
 
 public interface RoutePresenter {
+    companion object {
+        public val GESTURE_MIN_DELTA = 0.0001f
+    }
+
     public var routeController: RouteViewController?
     public var currentInstructionIndex: Int
 
     public fun onLocationChanged(location: Location)
     public fun onRouteStart(route: Route?)
     public fun onRouteResume(route: Route?)
-    public fun onMapGesture()
+    public fun onMapGesture(action: Int, pointerCount: Int, deltaX: Float, deltaY: Float)
     public fun onResumeButtonClick()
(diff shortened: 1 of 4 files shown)


app/src/main/kotlin/com/mapzen/erasermap/presenter/RoutePresenter.kt
```diff
@@ -6,13 +6,17 @@ import com.mapzen.valhalla.Instruction
 import com.mapzen.valhalla.Route
 
 public interface RoutePresenter {
+    companion object {
+        public val GESTURE_MIN_DELTA = 0.0001f
```
msmollin: How did you arrive at this number?

ecgreb: Its an arbitrary pixel value to make sure the user is actually trying to drag the map rather than perform another type of gesture.

msmollin: Is certain hardware able to track sub-pixel motion gestures? I hadn't heard of that. Just wondering if it needs to be that fine or not.

ecgreb: Good point. I'm not certain but [MotionEvent.html#getHistoricalX(int, int)](https://developer.android.com/reference/android/view/MotionEvent.html#getHistoricalX(int, int)) does return a float so I'm guessing its within the realm of posibility.

msmollin: Alrighty. Can always tweak later if we figure out something.

PR #201, opened 2015-12-15
Add settings to enable debug drawing for tiles and labels

### LINKED ISSUE #167: Add debug printout on display for map zoom level
It's hard to know what zoom level we're looking at to debug the scene file styling. Not a ship block, but a nice to have. Probably on the Tangram side to enable this, with an app settings toggle?

blair1618:
This is mostly to aid in style evaluation and debugging. With tile debug info enabled, the boundaries and coordinates (x, y, zoom) of each tile is made visible. With label debug info enabled, anchor points and bounding boxes for each label are made visible. 

I considered making this a single toggle but the label debug drawing is quite slow and usually not needed. Remaining nitpicks:

 - ~~Settings aren't applied until the settings menu is opened (i.e. until a `SettingsFragment` is created); not sure how to fix this most simply.~~ (Thanks Chuck!)
 - Tile debug info is only applied to new tiles, not already-rendered tiles. This is really a thing that's best fixed within Tangram-ES, so I'll take care of that there. 

Closes #167


ecgreb:
@blair1618 Tangram debug settings will now take effect on app launch.

tallytalwar:
This will be good.

PR #219, opened 2015-12-22
volume assets

souperneon:


ecgreb:
Added to https://github.com/mapzen/eraser-map/pull/363

PR #235, opened 2016-01-08
Peter/notifications

kingofirony:
- Adds routing notifications with instructions and ability to exit navigation.


ecgreb:
Do we have multiple versions of `ic_notif.png` for each density?

app/src/main/kotlin/com/mapzen/erasermap/util/NotificationBroadcastReceiver.kt
```diff
@@ -0,0 +1,32 @@
...
+        }
+    }
+
+    private fun cancelAllNotifications(context: Context) {
+        val notificationManager = context.getSystemService(Activity.NOTIFICATION_SERVICE) as NotificationManager
+        notificationManager.cancelAll()
+    }
+
+    private fun startBaseActivityWithExitExtra(context: Context) {
+        var exitRoutingIntent: Intent? = Intent(context, MainActivity::class.java)
```
ecgreb: Local value `exitRoutingIntent` can be declared immutable with `val` rather than `var`. Also no need to declare as nullable so type can be inferred.

    val exitRoutingIntent = Intent(context, MainActivity::class.java)

This will allow us to access via `.` below instead of `?.`

Also line ending semicolons can be dropped.

app/src/main/kotlin/com/mapzen/erasermap/view/RouteModeView.kt
```diff
@@ -114,6 +116,9 @@ public class RouteModeView : LinearLayout, RouteViewController, ViewPager.OnPage
                 routePresenter?.onInstructionSelected(instruction)
             }
         }
+        notificationCreator?.createNewNotification(
```
ecgreb: When testing the notification using a mock route, the instruction text shown in the notification always lagged one behind the current instruction shown in the view pager.

It seems to work better if this logic is moved into `setCurrentInstruction(index: Int)`

Also, `route?.getCurrentInstruction()` is not always the same as the current instruction that is displayed to the user (since we give the user advance warning of upcoming maneuvers). Better to use `route?.getRouteInstructions()?.get(index)` to determine which instruction to display.

ecgreb:
Notification should be dismissed when route is complete.

app/src/main/kotlin/com/mapzen/erasermap/util/NotificationBroadcastReceiver.kt
kingofirony: Makes sense - done

msmollin:
@kingofirony I'm having a little trouble following the comment thread here - did you resolve all of @ecgreb's comments? If so I'll merge it.

app/src/main/kotlin/com/mapzen/erasermap/presenter/MainPresenter.kt
```diff
@@ -43,4 +43,5 @@ public interface MainPresenter {
     public fun onReroute(location: Location)
     public fun onMapMotionEvent(): Boolean
     public fun onLongPressMap(event: MotionEvent): Boolean
+    open fun onExitNavigation()
```
blair1618: Is `open` redundant on an interface? My (limited) understanding is that `open` allows overriding, which is sort of built into an interface.

blair1618:
If I exit navigation from the notification when the app is not in the foreground, the app is then brought to the foreground. This is a little counter-intuitive to me; the user's intent is to dismiss the app but the result brings the app back to their attention. Is it easy to exit navigation without bringing the app back into the foreground?

app/src/main/kotlin/com/mapzen/erasermap/util/NotificationCreator.kt
```diff
@@ -0,0 +1,87 @@
...
+
+    private fun initExitNavigationIntent() {
+        exitNavigationIntent = Intent(mainActivity, NotificationBroadcastReceiver::class.java)
+        exitNavigationIntent?.putExtra(EXIT_NAVIGATION, true)
+        pendingExitNavigationIntent = PendingIntent.getBroadcast(
+                mainActivity, 0, exitNavigationIntent, PendingIntent.FLAG_CANCEL_CURRENT)
+    }
+
+    private fun initNotificationIntent() {
+        notificationIntent = Intent(mainActivity, MainActivity  ::class.java)
```
blair1618: Extra space after MainActivity

mjcunningham:
cc @souperneon for matt's comments on notification UX.

souperneon:
I agree with @blair1618 Exiting the navigation from the notification drawer should not open the app.
Also, what is the wording we are using for the action? "Exit Navigation"?

app/src/main/kotlin/com/mapzen/erasermap/util/NotificationCreator.kt
kingofirony: done

kingofirony:
@blair1618  @souperneon  Exit Navigation it is . And I'm on it - it results in some weird bug with  the map autocentering when you open the app next but it doesn't seem like it should take long to figure that out.

mjcunningham:
:+1: :+1:

app/src/main/kotlin/com/mapzen/erasermap/presenter/MainPresenterImpl.kt
```diff
@@ -300,8 +300,16 @@ public open class MainPresenterImpl(val mapzenLocation: MapzenLocation, val bus:
...
+    override fun onExitNavigation() {
+        vsm.viewState = ViewStateManager.ViewState.SEARCH_RESULTS
+        routingEnabled = false;
+        routeManager?.reverse = false
+        onFindMeButtonClick()
+    }
+
     override fun onMapMotionEvent(): Boolean {
         mainViewController?.rotateCompass()
+        onFindMeButtonClick()
```
blair1618: What's the intent of this call? The effect I see is that whenever I rotate the map the position is snapped back to my current location.

kingofirony: think I might have accidentally merged that in when I pulled

blair1618:
Getting there :)

Now if the app is currently in the foreground and I "Exit Navigation" from the notification, the app will be moved to the background. I would expect app foreground/background state to be unaffected by exiting navigation, maybe @souperneon has input here.

blair1618:
I'm also noticing some regressions with the "resume" behavior during navigation. 

 - The "resume" button is visible before I have moved the map away from the route
 - After I move the map away from the route, the map is snapped back to my current location (after a location update, I think?)

souperneon:
@blair1618 100% agree with the behavior you described. Exiting shouldn't influence background/foreground state

kingofirony:
@blair1618 Hmm I"m not having the same issue at all with the regressions on routing. Is the resume button just existent the second routing mode is entered? On another not the latest commit got the background/foreground state consistent.

blair1618:
App foreground/background state is looking good! I'm no longer noticing the position snapping during routing either, not sure what was up with that.  

The "resume" button still seems to have a small bug: If navigation is cancelled while the "resume" button is visible, it will be visible again immediately at the start of the next route. *However* now that I've checked again, I've found that this is not a regression, simply a bug in the current master. 

The new functionality looks correct to me and does not appear to introduce regressions, so I'll go ahead and merge! I can also create an issue for the "resume" button bug.

Issue #234, opened 2016-01-07
Enable tapping on map icons


nvkelso:
To pull up search style display with feature name and call to action to nav there.

nvkelso:
Related: https://github.com/mapzen/eraser-map/issues/233, which was previously said to be a blocker.

tallytalwar:
Is the expected behavior here, to "navigate to" the tapped icon?

PR #245, opened 2016-01-16
Select search icons and pois

### LINKED ISSUE #233: Enable tapping on search result pins
(Was there another issue for this, I don't see it!)

tallytalwar:
Fixes #233 and #234

### DIFF
--- a/app/src/main/kotlin/com/mapzen/erasermap/view/MainActivity.kt
+++ b/app/src/main/kotlin/com/mapzen/erasermap/view/MainActivity.kt
@@ -54,6 +54,15 @@ import javax.inject.Inject
 public class MainActivity : AppCompatActivity(), MainViewController, RouteCallback,
         SearchResultsView.OnSearchResultSelectedListener {
 
+    companion object {
+        @JvmStatic val MAP_DATA_PROP_TAP = "tap";
+        @JvmStatic val MAP_DATA_PROP_TAP_SEARCH = "search"
+        @JvmStatic val MAP_DATA_PROP_SEARCHINDEX = "searchIndex"
+        @JvmStatic val MAP_DATA_PROP_STATE = "state"
+        @JvmStatic val MAP_DATA_PROP_STATE_ACTIVE = "active"
+        @JvmStatic val MAP_DATA_PROP_STATE_INACTIVE = "inactive"
+    }
+
     public val requestCodeSearchResults: Int = 0x01
 
     var savedSearch: SavedSearch? = null
@@ -187,9 +196,9 @@ public class MainActivity : AppCompatActivity(), MainViewController, RouteCallba
         })
         mapController?.setFeatureTouchListener({
             properties ->
-                val tapProp = properties.getString("tap")
-                val searchIndexProp = properties.getNumber("searchIndex").toInt()
-                if (tapProp == "search") {
+                val tapProp = properties.getString(MAP_DATA_PROP_TAP)
+                val searchIndexProp = properties.getNumber(MAP_DATA_PROP_SEARCHINDEX).toInt()
+                if (tapProp == MAP_DATA_PROP_TAP_SEARCH) {
                     presenter?.onSearchResultTapped(searchIndexProp)
                 } else {
                     if (poiTapPoint != null) {
@@ -239,7 +248,6 @@ public class MainActivity : AppCompatActivity(), MainViewController, RouteCallba
     override fun showCurrentLocation(location: Location) {
         val currentLocation = LngLat(location.longitude, location.latitude)
         val properties = com.mapzen.tangram.Properties()
-        properties.add("type", "point");
 
         findMe?.clear()
         findMe?.addPoint(properties, currentLocation)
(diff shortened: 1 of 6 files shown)


nvkelso:
Got this installed on my phone over the weekend and it's working really good :) 1000% improvement!

Minor related issues I think with Pelias revGeo versus vector tiles icon placements versus tap versus returned latLng, and sometimes they return address instead of business (priority should be reversed). That shouldn't hold up this PR, let's talk about them in this week's mobile meeting.

tallytalwar:
Thanks for testing @nvkelso.

Another thing I noticed in some cases, geo coordinates returned from pelias for a poi vs what (poi) is being tapped was different.

app/src/main/kotlin/com/mapzen/erasermap/view/MainActivity.kt
```diff
@@ -38,6 +38,7 @@ import com.mapzen.pelias.widget.AutoCompleteListView
 import com.mapzen.pelias.widget.PeliasSearchView
 import com.mapzen.tangram.LngLat
 import com.mapzen.tangram.MapController
+import com.mapzen.tangram.MapController.FeatureTouchListener
```
ecgreb: Unused import

tallytalwar: Thanks, fixed.

app/src/main/kotlin/com/mapzen/erasermap/view/MainActivity.kt
```diff
@@ -408,21 +429,23 @@ public class MainActivity : AppCompatActivity(), MainViewController, RouteCallba
...
         }
 
-        var activeFeature: Int = 0
+        var featureCount: Int = 0
         searchResults?.clear()
         for (feature in features) {
             val simpleFeature = SimpleFeature.fromFeature(feature)
             val lngLat = LngLat(simpleFeature.lng(), simpleFeature.lat())
             val properties = com.mapzen.tangram.Properties()
+            properties.add("tap", "search")
```
ecgreb: We should consider adding constants for property values like `"tap"`, `"search"`, "`type`", and `"point"` so that we are not reusing the same string literals in several places.

Should these constants live in the app or does it make sense for some commonly used default values to be declared in Tangram?

app/src/main/kotlin/com/mapzen/erasermap/view/MainActivity.kt
```diff
@@ -439,6 +462,18 @@ public class MainActivity : AppCompatActivity(), MainViewController, RouteCallba
         }, 100)
     }
 
+    override fun centerOnTappedFeature(features: List<Feature>, position: Int) {
```
ecgreb: Perhaps the `centerOnCurrentFeature(features: List<Feature>)` method above could be refactored to simply call this method passing `pager.getCurrentItem()` as the `position` param?

Maybe also rename this method to `centerOnFeature(...)` to make it more generic?

```kotlin
    override fun centerOnCurrentFeature(features: List<Feature>) {
        val pager = findViewById(R.id.search_results) as SearchResultsView
        centerOnFeature(features, pager.getCurrentItem())
    }
```

app/src/main/kotlin/com/mapzen/erasermap/view/MainActivity.kt
tallytalwar: I think we can get rid of "type", "point" property here. Its added to the demo tangram app as an example usage of the property.

With respect to `"tap"`, `"search"` property name and value, these are specific values set in the app (in accordance with the usage in the style sheet). I don't think we need to add these to tangram library. 

Is there a specific file/styling convention you are following for constants, let me know and I can add these.

app/src/main/kotlin/com/mapzen/erasermap/view/MainActivity.kt
tallytalwar: Yes, makes sense. Will update this. Thanks

app/src/main/kotlin/com/mapzen/erasermap/view/MainActivity.kt
ecgreb: Great thanks!

See https://github.com/mapzen/eraser-map/blob/master/app/src/main/kotlin/com/mapzen/erasermap/view/InstructionListActivity.kt#L12 for an example of how to declare constants using a `companion object` and the `@JvmStatic` annotation.

tallytalwar: Thanks, updated. I went with a naming convention there ... "MAP_DATA_PROP_`PROPNAME`", "MAP_DATA_PROP_`PROPNAME`_`PROPVALUE`"

app/src/main/kotlin/com/mapzen/erasermap/view/MainActivity.kt
tallytalwar: Updated. Thanks.

PR #262, opened 2016-01-21
Style routing instructions

### LINKED ISSUE #211: Style current instruction view in routing mode
* Update turn icons
* Remove left and right pager control arrows (keeping functionality intact)
* Style distance and street name text
* This issue does not include top bar with total distance and time

(Image 1 - current instruction, Image 2 - previous or next instruction with resume button)

11a

image

ecgreb:
Style view pager and instruction icons in routing mode.

Closes #211 

image

image


app/src/main/kotlin/com/mapzen/erasermap/util/DisplayHelper.kt
```diff
@@ -3,15 +3,33 @@ package com.mapzen.erasermap.util
...
+     * Fetch the resource drawable ID of the large turn icon for the given instruction and style.
+     * @param context current context in which to display the icon.
+     * @param type the integer value representing this maneuver type.
+     * @return the resource ID of the turn icon to display.
+     */
+    public fun getRouteDrawableLarge(context: Context, type: Int?): Int {
+        return getRouteDrawableInternal(context, type, true)
+    }
+
+    private fun getRouteDrawableInternal(context: Context, type: Int?, isLarge: Boolean): Int {
```
msmollin: What's your reasoning for having separate functions for different size route drawables and not just have the code call the boolean-based function directly?

msmollin:
Can't comment on the images directly, but given the recent discussion to use blue lines instead of green, should we be pulling out all the blue icons?

mjcunningham:
@msmollin I say we move forward with the assets we have as they mess around with line color. Once they settle that Ekta can cut new icons and we can just pop them in.

msmollin:
Very well.

app/src/main/kotlin/com/mapzen/erasermap/util/DisplayHelper.kt
ecgreb: I didn't want to have to update all the other places we call this to add `false` as the third argument but.....

I am going to try rewriting using Kotlin's default argument syntax for `isLarge: Boolean = false` which might turn out to be cleaner.

https://kotlinlang.org/docs/reference/functions.html#default-arguments

ecgreb:
@msmollin I simplified the API for `DisplayHelper` using a default arg for `isLarge`. What do you think?

msmollin:
I like it! +1

PR #264, opened 2016-01-22
Multple feature picking updates

### LINKED ISSUE #254: On map feature select pin should point to icon not building centroid.
cc @blair1618  and @nvkelso for thoughts.

tallytalwar:
- Feature Picking:
    - If an osm id exists for a picked feature, Pelias Place Search is done to fetch the feature
    - Else reverseGeocode is performed at the tapped position
    - Feature Picked Pin location is grabbed from the centeroid of the picked feature (which is saved as a property in
    tangram). This makes sure feature pick pin is dropped at the correct position
- Small Updates to search pins picking (mainly brevity)

Fixes #254 

**IMPORTANT NOTE: This is dependent on https://github.com/tangrams/tangram-es/pull/489, and should not be merged before `tangram-es` PR is merged.**

### DIFF
--- a/app/src/main/kotlin/com/mapzen/erasermap/presenter/MainPresenterImpl.kt
+++ b/app/src/main/kotlin/com/mapzen/erasermap/presenter/MainPresenterImpl.kt
@@ -85,6 +85,7 @@ public open class MainPresenterImpl(val mapzenLocation: MapzenLocation, val bus:
             val current = searchResults?.features?.get(0)
             if (current is Feature) {
                 features.add(current)
+                mainViewController?.overridePlaceFeaturePosition(features?.get(0))
             }
             searchResults?.features = features
             mainViewController?.showPlaceSearchFeature(features)
--- a/app/src/main/kotlin/com/mapzen/erasermap/presenter/MainPresenterImpl.kt
+++ b/app/src/main/kotlin/com/mapzen/erasermap/presenter/MainPresenterImpl.kt
@@ -75,6 +75,22 @@ public open class MainPresenterImpl(val mapzenLocation: MapzenLocation, val bus:
         }
     }
 
+    override fun onPlaceSearchResultsAvailable(searchResults: Result?) {
+        vsm.viewState = ViewStateManager.ViewState.SEARCH_RESULTS
+        var features = ArrayList<Feature>()
+        this.searchResults = searchResults
+        if(searchResults?.features?.isEmpty() as Boolean) {
+            mainViewController?.emptyPlaceSearch()
+        } else {
+            val current = searchResults?.features?.get(0)
+            if (current is Feature) {
+                features.add(current)
+            }
+            searchResults?.features = features
+            mainViewController?.showPlaceSearchFeature(features)
+        }
+    }
+
     override fun onRestoreViewState() {
         if (destination != null) {
             if(routingEnabled) {
@@ -335,6 +351,15 @@ public open class MainPresenterImpl(val mapzenLocation: MapzenLocation, val bus:
     override fun onReverseGeoRequested(screenX: Float, screenY: Float): Boolean {
         if (reverseGeo || vsm.viewState == ViewStateManager.ViewState.DEFAULT) {
             mainViewController?.reverseGeolocate(screenX, screenY)
+            reverseGeo = true
+            return true
+        }
+        return false
+    }
+
+    override fun onPlaceSearchRequested(gid: String): Boolean {
+        if (reverseGeo || vsm.viewState == ViewStateManager.ViewState.DEFAULT) {
+            mainViewController?.placeSearch(gid)
             reverseGeo = true
             return true
         }
--- a/app/src/main/kotlin/com/mapzen/erasermap/presenter/MainPresenterImpl.kt
+++ b/app/src/main/kotlin/com/mapzen/erasermap/presenter/MainPresenterImpl.kt
@@ -349,11 +349,13 @@ public open class MainPresenterImpl(val mapzenLocation: MapzenLocation, val bus:
         return false
     }
 
-    override fun onReverseGeoRequested(screenX: Float, screenY: Float): Boolean {
-        if (reverseGeo || vsm.viewState == ViewStateManager.ViewState.DEFAULT) {
-            mainViewController?.reverseGeolocate(screenX, screenY)
-            reverseGeo = true
-            return true
+    override fun onReverseGeoRequested(screenX: Float?, screenY: Float?): Boolean {
+        if (screenX != null && screenY != null) {
+            if (reverseGeo || vsm.viewState == ViewStateManager.ViewState.DEFAULT) {
+                mainViewController?.reverseGeolocate(screenX, screenY)
+                reverseGeo = true
+                return true
+            }
         }
         return false
     }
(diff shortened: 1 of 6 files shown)


app/src/main/kotlin/com/mapzen/erasermap/presenter/MainPresenterImpl.kt
```diff
@@ -75,6 +75,23 @@ public open class MainPresenterImpl(val mapzenLocation: MapzenLocation, val bus:
         }
     }
 
+    override fun onPlaceSearchResultsAvailable(searchResults: Result?) {
```
ecgreb: Nice job using presenter to get this logic out of the activity :+1:

app/src/main/kotlin/com/mapzen/erasermap/view/MainActivity.kt
```diff
@@ -210,14 +209,21 @@ public class MainActivity : AppCompatActivity(), MainViewController, RouteCallba
...
+                    presenter?.onSearchResultTapped(searchIndex)
                 } else {
-                    if (poiTapPoint != null) {
-                        presenter?.onReverseGeoRequested(poiTapPoint?.get(0) as Float, poiTapPoint?.get(1) as Float)
+                    if (properties.contains(MAP_DATA_PROP_ID)) {
+                        val featureID = properties.getNumber(MAP_DATA_PROP_ID).toLong()
+                        presenter?.onPlaceSearchRequested("osm:venue:$featureID")
+                    } else {
+                        if (poiTapPoint != null) {
+                            presenter?.onReverseGeoRequested(poiTapPoint?.get(0) as Float, poiTapPoint?.get(1) as Float)
```
ecgreb: Using `as` to cast is somewhat dangerous as it will throw a runtime exception if the value is `null`.

In this case I think its fine since the `FeatureTouchListener` returns Java primitives for `positionX` and `positionY` they are guaranteed to be non-null values within the scope of this function.

app/src/main/kotlin/com/mapzen/erasermap/view/MainActivity.kt
```diff
@@ -434,13 +453,25 @@ public class MainActivity : AppCompatActivity(), MainViewController, RouteCallba
...
-        val simpleFeature = SimpleFeature.fromFeature(features.get(0))
-        val lngLat = LngLat(simpleFeature.lng(), simpleFeature.lat())
+        var lngLat: LngLat?
+
+        if (poiTapPoint == null) {
+            val simpleFeature = SimpleFeature.fromFeature(features.get(0))
+            lngLat = LngLat(simpleFeature.lng(), simpleFeature.lat())
+        } else {
+            val point = poiTapPoint
+            lngLat = mapController?.coordinatesAtScreenPosition(point!![0].toDouble(), point!![1].toDouble())
```
ecgreb: Using `!!` is also kind of scary since its one of the few ways to get a null pointer exception in Kotlin. Something like this would be safer (but would also require updating line 467 below since `lngLat` now may not have been initialized).

```kotlin
            val lng = poiTapPoint?.get(0)?.toDouble()
            val lat = poiTapPoint?.get(1)?.toDouble()
            if (lng != null && lat != null) {
                lngLat = mapController?.coordinatesAtScreenPosition(lng, lat)
            }
```

app/src/main/kotlin/com/mapzen/erasermap/view/MainActivity.kt
```diff
@@ -769,6 +811,19 @@ public class MainActivity : AppCompatActivity(), MainViewController, RouteCallba
         hideReverseGeolocateResult()
     }
 
+    override fun overridePlaceFeaturePosition(feature: Feature) {
+        if (poiTapPoint != null) {
+            val geometry = Geometry()
+            val coordinates = ArrayList<Double>()
+            var coords = mapController?.coordinatesAtScreenPosition(
```
ecgreb: Again, maybe there is a way to rewrite this section with less `!!`. Its not super critical since we control the Tangram library and trust it. But if Tangram was a third party product I would never want to use `!!` on nullable return values ;)

app/src/test/kotlin/com/mapzen/erasermap/presenter/MainPresenterTest.kt
```diff
@@ -62,6 +62,14 @@ public class MainPresenterTest {
         assertThat(mainController.isReverseGeocodeVisible).isTrue()
     }
 
+    @Test fun onPlaceSearchResultsAvailable_shouldShowSearchResults() {
```
ecgreb: Can we add a test case to cover the code path when `features` is not empty?

app/src/main/kotlin/com/mapzen/erasermap/view/MainActivity.kt
tallytalwar: It was failing to compile, because its expecting a `Float` and it found a `Float?`, so I added that `as`. Good to know the usage though.

Do you suggest another way to doing this? Checking if `poiTapPoint?.get(0)` is not null and they grabbing value from there?

ecgreb: In this case I might have chosen to make the params type `Float?` and then do input validation in the presenter. And yes, in general `?.` and `toFloat()` are usually preferable to `!!.` or `as`.

There is also `as?` for null-safe casting but I haven't found a lot of places where this is useful so far.

tallytalwar: Thanks.

Addressed by: 18371ee

app/src/main/kotlin/com/mapzen/erasermap/view/MainActivity.kt
tallytalwar: Addressed by: 18371ee

app/src/main/kotlin/com/mapzen/erasermap/view/MainActivity.kt
tallytalwar: Addressed by: 18371ee

app/src/test/kotlin/com/mapzen/erasermap/presenter/MainPresenterTest.kt
tallytalwar: Addressed by: 1e8b6a7

mjcunningham:
Inday on our block shows up with an icon and a label as "Inday"  but when I tap it I get the address. Is that expected?

tallytalwar:
@mjcunningham yes when a pelias place search results nothing, code falls back to reverseGeo on the tapped location.

A bit more explanation:
- tapped poi's osm id is extracted to form a global id, which at present is: "osm:venue:`poi_osm_id`".
- pelias place search is performed with this global id.
- if pelias place search returns a feature, you get appropriate place name.
- if pelias place search returns no feature(s), it falls back to do a reverseGeo at the tapped position which will return an address!

I think this will also get better when our tiles have features from more sources like wof, which will result in better pelias place search matches.

tallytalwar:
Pelias Result for "osm:venue:3971304160": https://search.mapzen.com/v1/place?ids=osm%3Avenue%3A3971304160&api_key=search-hHoQt3U (Inday's OSM ID: `3971304160`)

tallytalwar:
@mjcunningham I might be able to do something there, can you file an `enhancement` issue for the same? Thanks

PR #282, opened 2016-01-27
Add debug toggle for misc tangram debug infos

karimnaaji:
Precise zoom display requested from https://github.com/mapzen/eraser-map/issues/167 with more misc tangram infos printout (current lat/lon, tile cache size, visible tiles, frame time, ...) from https://github.com/tangrams/tangram-es/pull/493


ecgreb:
:tada:

PR #288, opened 2016-01-29
new assets for nav screen

souperneon:
@ecgreb sorry for being so late… but I  made both normal and pressed states… went one resolution size smaller. See if this works!


ecgreb:
@souperneon looks like we are missing ic_map at the new resolution and pressed state?

PR #289, opened 2016-01-30
Route mode view updates part 2

### LINKED ISSUE #88: Update route list view while in route (active cell)
Phone: Samsung Galaxy S5
Android Version: 5.0

First of all I like that even in this early version we have the option of checking the route list while in route/navigation mode. Second thing I really like is that its a transparent view, and the underneath map/route is visible.

It would be good to position the route list view to the current already traveled position rather than showing the route list from the very beginning of the route.

In the screen shot below, I was on middle of a route on I-95 North bound, and when I opened route list, it started from the very beginning (from my apartment). I usually use route list in navigation to see whats coming next.

routelist

### LINKED ISSUE #212: Move total route distance and time view from bottom to top
* Replace route footer view with new top bar above directions
* Replace sliding drawer with icon to open list view
* Add drop-down list to view all instructions
* Add exit route "X" icon

image

image

ecgreb:
* Move distance and destination view from footer to above instruction pager (#212)
* Remove sliding up panel and implement new direction list UI (#212)
* Adjust layout for configuration changes and landscape mode (#212)
* Highlight current instruction in direction list view (#88)
* Adds new routing icons (#214 and #288 )

Closes #212 
Closes #88 
Closes #214 
Closes #288 

image

image


ecgreb:
@msmollin @kingofirony @tallytalwar I am going to merge this sans review so its available for weekend driving tests. Please post any comments when you get a chance and I'll address them in a separate PR!

PR #295, opened 2016-02-01
Restore view state based on VSM

### LINKED ISSUE #276: Hitting back in route preview should clear the route
Repro:

- Get into route preview
- Back all the way until the app closes
- Re-open the app
- You are trapped

ecgreb:
Revises logic in `MainPresenter#onRestoreViewState()` to use `ViewStateManager` to dictate which views should be shown and/or hidden when the activity is recreated. Prevents getting stuck in route preview mode when resuming the app once and for all (I hope).

Tacked on some code cleanup including static import of view states and Kotlin property access syntax.

Fixes #276

### DIFF
--- a/app/src/main/kotlin/com/mapzen/erasermap/presenter/MainPresenterImpl.kt
+++ b/app/src/main/kotlin/com/mapzen/erasermap/presenter/MainPresenterImpl.kt
@@ -9,6 +9,12 @@ import com.mapzen.erasermap.model.event.LocationChangeEvent
 import com.mapzen.erasermap.model.event.RouteCancelEvent
 import com.mapzen.erasermap.model.event.RouteEvent
 import com.mapzen.erasermap.model.event.RoutePreviewEvent
+import com.mapzen.erasermap.presenter.ViewStateManager.ViewState.DEFAULT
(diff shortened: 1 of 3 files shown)


tallytalwar:
Looks good to me and fixes the issue :D.

app/src/main/kotlin/com/mapzen/erasermap/presenter/MainPresenterImpl.kt
```diff
@@ -94,23 +100,43 @@ public open class MainPresenterImpl(val mapzenLocation: MapzenLocation, val bus:
...
+            SEARCH -> onRestoreViewStateSearch()
+            SEARCH_RESULTS -> onRestoreViewStateSearchResults()
+            ROUTE_PREVIEW -> onRestoreViewStateRoutePreview()
+            ROUTING -> onRestoreViewStateRouting()
+            ROUTE_DIRECTION_LIST -> onRestoreViewStateRouteDirectionList()
         }
+   }
+
+    private fun onRestoreViewStateDefault() {
+        // Do nothing.
```
tallytalwar: I was thinking of calling initFindMeButton here!!... But while I am writing this, I do not think its a good idea, since that will always bring you back to your current location instead of restoring the previous last known state. :D.

PR #306, opened 2016-02-02
Be more aggressive in clearing gl resources

tallytalwar:
- This will make tangram clear its gl resources whenever the app is backgrounded.
- Possibly will not make android kill eraser-map activity that often when other applications are fighting for resources.

cc. @blair1618 @ecgreb

### DIFF
--- a/app/src/main/kotlin/com/mapzen/erasermap/view/MainActivity.kt
+++ b/app/src/main/kotlin/com/mapzen/erasermap/view/MainActivity.kt
@@ -186,6 +186,7 @@ public class MainActivity : AppCompatActivity(), MainViewController, RouteCallba
 
     private fun initMapController() {
         val mapView = findViewById(R.id.map) as MapView
+        mapView.setPreserveEGLContextOnPause(false)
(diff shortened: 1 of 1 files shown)


blair1618:
I'm a little skeptical that this will improve anything. It sounds like the hypothesis here is that Android is killing the app because it needs to reclaim some GL resources, but Activity lifecycle management is pretty independent from GL resource management. If tangram's GL resources are constraining another application, Android will destroy our GL context whether or not we request the context to be preserved on pause (which we do by default). However if we request that the context *not* be preserved on pause (i.e. this PR) then the app will always have wait for GL resources to be rebuilt on resume, even if they could have been kept.

tallytalwar:
I agree on what you said there, but wanted to give this a shot considering the instances I have seen the app being destroyed while in routing.

However, recent knowledge, I gained from reading [this](https://medium.com/google-developers/who-lives-and-who-dies-process-priorities-on-android-cb151f39044f#.b4uyno8ii) article, which @ecgreb sent, makes sense and hence probably this might not do anything!

PR #307, opened 2016-02-03
Replace splash screen animation with static image

### LINKED ISSUE #107: Splash screen animation cut off
Samsung Galaxy S2
Android 4.0.4

ezgif com-crop

ecgreb:
* Replaces splash screen animation with a single static asset.
* Adds 3 second fixed duration to display splash screen.

TODO: Logo text does not scale up well on larger devices (Nexus 9) so we may want to replace single asset with separate logo and background images.

Closes #107
Closes #281


app/src/main/kotlin/com/mapzen/erasermap/view/InitActivity.kt
```diff
@@ -12,70 +10,28 @@ import com.mapzen.erasermap.BuildConfig
...
-                }
-                if(letterIndex == 5) {
-                    startMainActivity()
-                }
-            }
-        })
-        bottomAnimation.start()
-    }
-
-    public fun startMainActivity() {
```
tallytalwar: just curious, why was it public before?

ecgreb: To expose for testing

PR #309, opened 2016-02-03
Route Preview Fixes

### LINKED ISSUE #287: Route on screen doesn't clear when no route is found
Steps:
1. search for Rochester, NY (when in the office)
2. Select car in choose mode
3. Select walk
4. No route found but bike route still drawn on the screen

### LINKED ISSUE #298: "No route found" messes up the View State
Trying to a route to a place for which "no route is found", sets the view state to `ROUTE_PREVIEW`, but it should actually back up and set the view state to previous state, `SEARCH_RESULTS`!

tallytalwar:
- Show route preview with no route when no route is found.
        - Still draw the start and destination pins!
        - Reset Distance and Time to destination
- Refactored some code around drawing routing pins

Future work: Adjust map view to fit the 2 routing pins in screen when no route is found.

Fixes #298
Fixes #287

### DIFF
--- a/app/src/main/kotlin/com/mapzen/erasermap/view/TimeView.kt
+++ b/app/src/main/kotlin/com/mapzen/erasermap/view/TimeView.kt
@@ -11,11 +11,18 @@ import com.mapzen.erasermap.R
 public class TimeView(context: Context, attrs: AttributeSet) : TextView(context, attrs) {
     public var timeInMinutes: Int = 0
         set (value) {
-            if (value < 1) {
+            if (noTime) {
+                text = ""
+            } else if (value < 1) {
                 text = context.getString(R.string.less_than_one_minute)
             } else {
                 formatTime(value)
             }
+        }
+
+    public var noTime: Boolean = false
+        set (value) {
+            field = value
         }
 
     private fun formatTime(value: Int) {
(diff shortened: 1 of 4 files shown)


app/src/main/kotlin/com/mapzen/erasermap/view/TimeView.kt
```diff
@@ -11,13 +11,20 @@ import com.mapzen.erasermap.R
...
+            if (noTime) {
+                text = ""
+            } else if (value < 1) {
                 text = context.getString(R.string.less_than_one_minute)
             } else {
                 formatTime(value)
             }
         }
 
+    public var noTime: Boolean = false
```
tallytalwar: NOTE: Did not know how to handle "time = 0" case when no route was found, hence this `noTime` thingy!.

ecgreb:
Currently divider "|" character is still shown even when destination and time views are hidden. To fix this (and make adding a property to `TimeView` unnecessary) I would try assigning an id to the `LinearLayout` that wraps the destination and time.

https://github.com/mapzen/eraser-map/blob/master/app/src/main/res/layout/view_route_preview.xml#L33

Then you could use `findViewById(...)` to set the visibility of this entire view all at once when a route success fails or succeeds.

tallytalwar:
@ecgreb Did you mean something like this: 9ac4d91?
learning++ :D.

ecgreb:
Exactly!

PR #317, opened 2016-02-04
Custom search view

### LINKED ISSUE #299: Make search bar go full width on tablet

ecgreb:
* Removes search view menu item with action view auto-inflated by options menu.
* Adds `PeliasSearchView` to action bar as fully custom view with layout params.
* Updates custom search icon and clear icon.
* Search view now expands to full width of screen (minus action icons) in portrait and landscape mode.

Fixes #299 

image

### DIFF
--- a/app/src/main/kotlin/com/mapzen/erasermap/view/MainActivity.kt
+++ b/app/src/main/kotlin/com/mapzen/erasermap/view/MainActivity.kt
@@ -6,6 +6,7 @@ import android.location.Location
 import android.os.Bundle
 import android.os.Handler
 import android.preference.PreferenceManager
+import android.support.v7.app.ActionBar
 import android.support.v7.app.AppCompatActivity
 import android.util.Log
 import android.view.Menu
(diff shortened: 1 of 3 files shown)


tallytalwar:
Code looks good to me.

Question?
What should happen if I do the following:
1. Type in something for search: say "park"
2. Now instead of pressing `Done` in the keyboard, I press the **back** button?

Currently this take me to the default view (as expected), but the search action bar still has the word "park", and :heavy_multiplication_x: icon. (Refer the screen shot). And on pressing the :heavy_multiplication_x:, it takes me to the search list!

This makes me think, that pressing back on an incomplete search should also clear the search action bar? Thoughts?

screenshot_2016-02-04-19-01-33

mjcunningham:
I agree w @tallytalwar but cc-ing @souperneon for a vote.

ecgreb:
What the system back button should do and what the search cancel "X" should do is really two separate things.

1. I agree the back button should clear the current query text when used to close search to prevent the situation @tallytalwar is describing. I'll push an update for this.

2. The "X" cancel button should only every clear the current query but not exit search mode. Recent searches should still be displayed in the list view and the keyboard should remain open.

@souperneon does this match you understanding?

tallytalwar:
I agree on the "X" functionality. My comment was for handling system back button :D.

souperneon:
100% @ecgreb :+1:

souperneon:
@tallytalwar nice catch!

ecgreb:
Fixing the search box back button behavior is going to require a chance in the Pelias SDK. Added an issue here https://github.com/pelias/pelias-android-sdk/issues/30

PR #325, opened 2016-02-05
Update route icon positioning

### LINKED ISSUE #323: Move route arrow lower on the screen
We can use some of that free space now that the bottom panel is gone.

tallytalwar:
- sets the tilt angle to 47.5 (from 60).. since now we have perspective camera we do not need that much tilt
- update the position of the route icon at quater of the screen!

Fixes #323 

It looks like this now:

screenshot_2016-02-05-12-13-56

screenshot_2016-02-05-12-14-05


tallytalwar:
Also ignore the left and right direction mixup! (there is already an issue for that :D)

souperneon:
thanks @tallytalwar

PR #329, opened 2016-02-05
splash screen and map icon

souperneon:
- removed shadow from splash screen
- resized map icon


ecgreb:
If the splash screen background is just transparent now it can be removed from the asset and the icon can simply be centered on the screen.

PR #340, opened 2016-02-08
Update README with more specific developer api key information

### LINKED ISSUE #337: README update

sarahlensing:
-Rearranged the order of instructions so that the developer doesn't try building before necessary dependencies are installed

-Updated api key information with link to developer site

Closes #337


ecgreb:
:+1:

PR #359, opened 2016-02-09
Route preview zooms to fit bounds of the route

blair1618:
When previewing a route, this will calculate the lon/lat bounding box of the route geometry and move/zoom the map to show the entire route, taking into account the current screen dimensions, with a configurable amount of padding. If no route is found for a search, the preview will now zoom to fit the start and end pins. 

This also adds a utility class for an axis-aligned bounding box - might be useful in other places.

### DIFF
--- a/app/src/main/kotlin/com/mapzen/erasermap/util/AxisAlignedBoundingBox.kt
+++ b/app/src/main/kotlin/com/mapzen/erasermap/util/AxisAlignedBoundingBox.kt
@@ -1,0 +1,65 @@
+package com.mapzen.erasermap.util
+
+/**
+ * An axis-aligned bounding box
+ *
+ * Useful class for finding the axis-aligned bounds of a set of 2D points
+ *
+ * @author Matt Blair
+ */
+
+public class AxisAlignedBoundingBox {
+
+    /**
+     * 2D point container like PointF, but with double precision
+     */
+    public class PointD(x: Double = 0.0, y: Double = 0.0) {
+        var x: Double = x
+        var y: Double = y
+    }
+
+    var min = PointD()
+    var max = PointD()
+
+    var center: PointD
+        get() {
+            return PointD(0.5 * (max.x + min.x), 0.5 * (max.y + min.y))
+        }
+        set(c: PointD) {
+            val hw = 0.5 * this.width
+            val hh = 0.5 * this.height
+            max = PointD(c.x + hw, c.y + hh)
+            min = PointD(c.x - hw, c.y - hh)
+        }
+
+    var width: Double
+        get() {
+            return max.x - min.x
+        }
+        set(w: Double) {
+            val c = this.center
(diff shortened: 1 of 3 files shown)


tallytalwar:
Looks like, its does not zoom out the map when no route is found (in all the cases).

Working repro:
Search/RouteTo: Paris, France

Not Working Repro:
Search/RouteTo: Auckland, NZ

blair1618:
Aw dang, the bounding box is probably wrong in some way when the "route" crosses the international dateline. I'll look into that.

tallytalwar:
yeah, I think you will have to wrap the view somehow.

blair1618:
Hm ok, so all of the logic in this PR is handling these cases correctly. The unusual behavior actually comes from a bit of code in tangram-es that constrains the view from extending beyond the edges of the world. 

When you try to set a view parameter that would result in seeing area beyond the edge of the world, tangram-es stops the change completely and remains at the previous value. In this case, when we try to zoom out and see both New York and Auckland, the resulting zoom level would make visible some area beyond the edge of the world, so the map remains at the previous zoom value (probably 16). What I would expect as a developer and as a user would be for tangram-es to take the closest possible value to the input, rather than doing nothing. 

I can file an issue for tangram-es to improve this behavior, but for now it might be best to not attempt a route preview when no route is found.

tallytalwar:
Yes mostly thats the case.

Another case which I tried is "Delhi, India". Both "New York" and "Delhi" do fit the tangram view at zoom:1, but route preview fails to show this.

screenshot_2016-02-09-19-02-18

blair1618:
For the case of Delhi, the app is still trying to zoom out farther than tangram-es will allow; the route preview zooms to fit a box containing both endpoints plus 35% padding.

tallytalwar:
Yup totally.

I agree on your suggestion of handing this on tangram-es. This pr looks
good to me with respect to bounding box logic.

mjcunningham:
Is the behavior in this PR that if no route is found you cant get into the preview view?

blair1618:
With the changes in this PR, when no route is found the "preview" will move
and zoom to show the start and end locations. There are some exceptional
cases that Varum pointed out where the zoom will not be correct for
extremely long (read, intercontinental) routes.

tallytalwar:
@mjcunningham Nopes, it only takes care of zoom out to fit the route (if one is found), and to fit the start and end pins (if no route is found).

However on checking other mapping apps, when no route is found, they just show the last map view without trying to fit the start and end markers in the same map view (not possible all the time, as discussed above)

mjcunningham:
:+1: lets try this and see how it feels.

blair1618:
The last commit addresses these uncommon cases where tangram can't set the desired zoom - instead the map will just stay at the current location.

tallytalwar:
This looks good to me.

PR #368, opened 2016-02-12
Route complete

### LINKED ISSUE #154: Routing has completion issues
* **Device name:** Samsung Galaxy S6 edge
* **Android Version:** 5.1.1
* **App build number:** master-481
* **What did you expected to happen?** When I'm near my destination, I should see the final route instruction. 
* **What happened instead?** I only ever see the next to last route maneuver UI cell contents. If I manually swipe over to see the final one, it has a red background.
* **Steps to reproduce:** Simulate location to SF. Search for "eureka california".
* **Attach a screenshot** See below.
* **Attach device logs** eh.

If something happened while you were routing, share with us:

* **Where were you?** Target Eureka, California
* **Routing from origin:** San Francisco
* **Routing to destination:** Target Eureka, California

ecgreb:
Update UI at destination so it looks like route is actually complete.

* Display final instruction.
* Hide instruction distance view.
* Hide distance to destination.
* Set route icon location and center map.
* Show top bar with destination name and exit button.

Closes #154 

image

### DIFF
--- a/app/src/main/kotlin/com/mapzen/erasermap/view/InstructionAdapter.kt
+++ b/app/src/main/kotlin/com/mapzen/erasermap/view/InstructionAdapter.kt
@@ -11,8 +11,8 @@ import com.mapzen.erasermap.util.DisplayHelper
 import com.mapzen.valhalla.Instruction
 import java.util.ArrayList
 
-public class InstructionAdapter(val context: Context, val instructions: ArrayList<Instruction>,
-        val pager: RouteModeView) : PagerAdapter() {
+public class InstructionAdapter(val context: Context,
+        val instructions: ArrayList<Instruction>) : PagerAdapter() {
 
     override fun instantiateItem(container: ViewGroup?, position: Int): Any? {
         val instruction = instructions.get(position)
--- a/app/src/main/kotlin/com/mapzen/erasermap/view/InstructionAdapter.kt
+++ b/app/src/main/kotlin/com/mapzen/erasermap/view/InstructionAdapter.kt
@@ -11,8 +11,8 @@ import com.mapzen.erasermap.util.DisplayHelper
 import com.mapzen.valhalla.Instruction
 import java.util.ArrayList
 
-public class InstructionAdapter(val context: Context,
-        val instructions: ArrayList<Instruction>) : PagerAdapter() {
+public class InstructionAdapter(private val context: Context,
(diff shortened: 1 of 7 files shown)


app/src/main/kotlin/com/mapzen/erasermap/view/InstructionAdapter.kt
```diff
@@ -11,8 +11,8 @@ import com.mapzen.erasermap.util.DisplayHelper
 import com.mapzen.valhalla.Instruction
 import java.util.ArrayList
 
-public class InstructionAdapter(val context: Context, val instructions: ArrayList<Instruction>,
-        val pager: RouteModeView) : PagerAdapter() {
+public class InstructionAdapter(private val context: Context,
```
sarahlensing: Excuse the kotlin-101 question but what does making theses arguments private do? Are they public by default?

app/src/main/kotlin/com/mapzen/erasermap/view/RouteModeView.kt
```diff
@@ -319,9 +323,17 @@ public class RouteModeView : LinearLayout, RouteViewController, ViewPager.OnPage
     }
 
     override fun showRouteComplete() {
-        findViewById(R.id.footer_wrapper)?.visibility = View.GONE
         findViewById(R.id.resume)?.visibility = View.GONE
+        setCurrentInstruction(pager?.adapter?.count?.minus(1) ?: 0)
         notificationCreator?.killNotification()
+        (findViewById(R.id.destination_distance) as DistanceView).distanceInMeters = 0
```
sarahlensing: Why not create a reference to this so the code is more concise? It seems like we reference this id more than once

app/src/main/kotlin/com/mapzen/erasermap/view/InstructionAdapter.kt
ecgreb: Kotlin automatically converts constructor arguments to properties if you put `val` or `var` in front and default scope is package private. Just wanted to update these so they are not visible outside the class.

app/src/main/kotlin/com/mapzen/erasermap/view/RouteModeView.kt
ecgreb: Good call. Added lazy val for distance to destination as well as view pager and resume button.

PR #370, opened 2016-02-12
Add osm attribution text to map

### LINKED ISSUE #117: Missing attribution
* Device name: Samsung Galaxy Tab 4 8.0 16GB, Black
* Android version: 4.4.4
* App build number: master-442
* What did you expect to happen? I expected to see attribution, either prominently in the splash screen or visible on the map, explaining that some of the data comes from OpenStreetMap.
* What happened instead? There's no attribution.
* Steps to reproduce: Open the app.

When using OpenStreetMap data, the requirement to attribute is not optional. We should include [the required credit](http://www.openstreetmap.org/copyright) either on the map, or prominently in the splash screen.

sarahlensing:
Adds TextViews to:
view_route_mode.xml
view_route_preview.xml
activity_main.xml

For activity_main.xml we need to dynamically update the LayoutParams when search results show/hide. I'm updating the values of layout_alignParentBottom and layout_above in code. I could have also implemented by adjusting the layout_marginBottom but it felt less descriptive.

Closes #117

### DIFF
--- a/app/src/main/kotlin/com/mapzen/erasermap/view/MainActivity.kt
+++ b/app/src/main/kotlin/com/mapzen/erasermap/view/MainActivity.kt
@@ -16,6 +16,8 @@ import android.view.inputmethod.EditorInfo
 import android.widget.Button
 import android.widget.ImageButton
 import android.widget.RadioButton
+import android.widget.RelativeLayout
+import android.widget.TextView
 import android.widget.Toast
 import com.mapzen.erasermap.BuildConfig
 import com.mapzen.erasermap.CrashReportService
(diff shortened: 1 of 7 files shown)


ecgreb:
Looks like the build is failing due to some missing imports.

ecgreb:
This is a super knit pick but when there are multiple search results the attribution (and find me button) get pushed up in a weird way due to the "1 of 10 results" text. It would be great if they were in the same position for single and multiple results.

***Single Result***

image

***Multiple Results***

image

sarahlensing:
Imports are fixed and we are now we are handling positioning well for multiple results. I implemented using bottom margin

PR #371, opened 2016-02-15
Peter/search list style

### LINKED ISSUE #209: Style autocomplete list view
image

kingofirony:
- Opens SearchView with list of locations in the pager when view all is clicked. 
- Adds New Search Style
screenshot_2016-02-15-05-42-15

Closes #209


ecgreb:
I am seeing some behavior that is inconsistent with the screenshot above and #209. (Nexus 5, Android 6.0)

1. Address text appears on same line as name then wraps to next line **(Screenshot 1)**
2. Long place names should ellipsize rather than wrap (maybe slightly smaller font better here).  **(Screenshot 1)**
3. After completing search query text is cleared (should stay). **(Screenshot 2)**
4. After completing search recent search (timer) icon is used instead of result (pin) icon. **(Screenshot 2)**
5. Using system back button to close search results list view does not update action bar icon. **(Screenshot 3)**

**Screenshot 1**
image

**Screenshot 2**
image

**Screenshot 3**
image

sarahlensing:
I'm also seeing a couple of the same issues:

1. Long names are not ellipsizing
2. Recent searches are not showing the recent icon

ecgreb:
This is looking better but it seems the UX is still not quite right. Clicking on an item in the search results list when there are multiple results clears all other results and only displays the result selected. Instead it should zoom to that result but keep others in list and view pager.

Steps to reproduce:
1. Launch app and tap search view
2. Enter "pizza" and tap enter key on soft keyboard to execute search
3. Tap action bar "list" icon to view all results
4. Tap second item in the search results list
5. Notice now there is only a single result in the pager. Also map/list toggle icon is gone.

This is the proper behavior when selecting an item from autocomplete but not when selecting an item from the search results list view.

kingofirony:
Build will fail until this is merged in. 
https://github.com/pelias/pelias-android-sdk/pull/31

PR #379, opened 2016-02-17
Query for special search term to reveal debug settings

### LINKED ISSUE #352: Hide debug settings in production builds

sarahlensing:
Searching for 8 exclamation points will reveal debug settings

Closes #352


ecgreb:
Very nice. After testing this out I have a couple additional requests that I think would make this feature even nicer:

1. Searching for secret phrase the first time will display a toast message to alert the user debug settings have been enabled.
2. Once debug settings are enabled searching for the secret phrase again will re-hide them. Also a toast message will be displayed to indicate debug mode is off.

app/src/main/kotlin/com/mapzen/erasermap/util/StringConstants.kt
```diff
@@ -0,0 +1,15 @@
+package com.mapzen.erasermap.util
+
+/**
+ *  Declaration for all string constants to be used in the app
+ */
+public class StringConstants {
```
ecgreb: This is a matter of style but in general I'm not a big fan of an app-wide class for constants that don't necessarily relate to each other.

Seems like these values would be better served in the class to which they are relevant. In this case maybe we could add them to the `AppSettings` interface?

sarahlensing:
I like these requests. I'll add them shortly.

msmollin:
Just a thought - would ~ be a better character for this? I'm concerned that ! is common enough that people will stumble upon these.

sarahlensing:
I agree that ! is a more common character than ~ but I think the likelihood that they will type 8 !'s in is about the same as 8 ~'s

sarahlensing:
Toast and toggle functionality added

ecgreb:
@sarahlensing updates look good. Looks like you still have some conflicts with master that need to be resolved though.

Also would like to get your thoughts about my comment on `StringConstants`. What do you think about moving those constants to `AppSettings` interface or possibly `AndroidAppSettings` class?

sarahlensing:
@ecgreb I think moving them to `AndroidAppSettings` makes sense, I'll do that and also resolve the conflicts with master now.

blair1618:
I formally request a konami code easter egg.

msmollin:
I formally :+1: @blair1618's formal request.

PR #392, opened 2016-02-19
new icons

souperneon:


ecgreb:
@souperneon ic_route_1 (start), ic_route_2 (start right), and ic_route_3 (start left) should all be a straight arrow so it is not confused with an actual right or left turn.

PR #443, opened 2016-03-02
tangram moved to okhttp3

tallytalwar:
- updates TileHttpHandler


ecgreb:
Since this somehow seems to be linked to CI issues let's hold off until after we get a stable build for launch.

PR #445, opened 2016-03-02
Add toggle for caching searches

### LINKED ISSUE #388: Add "cache nothing" toggle to settings
This should just be for search results and under the Privacy section.

 

- [ ] change "Erase History" to "Erase Search History" 

- [ ] Toggle should say "Cache Search History" and be set to on by default.

sarahlensing:
Dependent on: https://github.com/pelias/pelias-android-sdk/pull/36

Closes #388

### DIFF
--- a/app/src/main/kotlin/com/mapzen/erasermap/view/MainActivity.kt
+++ b/app/src/main/kotlin/com/mapzen/erasermap/view/MainActivity.kt
@@ -62,8 +62,7 @@ import java.util.ArrayList
 import javax.inject.Inject
 
 public class MainActivity : AppCompatActivity(), MainViewController, RouteCallback,
-        SearchResultsView.OnSearchResultSelectedListener,
-        SharedPreferences.OnSharedPreferenceChangeListener {
+        SearchResultsView.OnSearchResultSelectedListener {
 
     companion object {
         @JvmStatic val MAP_DATA_PROP_SEARCHINDEX = "searchIndex"
@@ -147,7 +146,6 @@ public class MainActivity : AppCompatActivity(), MainViewController, RouteCallba
         settings?.initTangramDebugFlags()
         initVoiceNavigationController()
         initNotificationCreator()
-        initPreferencesListener()
     }
 
     override protected fun onNewIntent(intent: Intent?) {
@@ -186,6 +184,7 @@ public class MainActivity : AppCompatActivity(), MainViewController, RouteCallba
         app?.onActivityResume()
         autoCompleteAdapter?.clear()
         autoCompleteAdapter?.notifyDataSetChanged()
+        invalidateOptionsMenu()
     }
 
     override public fun onPause() {
@@ -207,8 +206,6 @@ public class MainActivity : AppCompatActivity(), MainViewController, RouteCallba
         saveCurrentSearchTerm()
         routeModeView.clearRoute()
         findMe?.clear()
-        val preferences = PreferenceManager.getDefaultSharedPreferences(this)
-        preferences.unregisterOnSharedPreferenceChangeListener(this)
     }
 
     private fun initMapController() {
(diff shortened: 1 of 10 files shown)


app/src/main/kotlin/com/mapzen/erasermap/view/MainActivity.kt
```diff
@@ -1057,6 +1068,13 @@ public class MainActivity : AppCompatActivity(), MainViewController, RouteCallba
         poiTapPoint = null
     }
 
+    override fun onSharedPreferenceChanged(preferences: SharedPreferences?, key: String?) {
```
ecgreb: Do we really need a preference listener here? By definition `MainActivity` will be in the background whenever this value changes. Would it suffice to simply update the value in `onResume()` or maybe better yet `onPrepareOptionsMenu()`?

app/src/main/kotlin/com/mapzen/erasermap/view/SettingsFragment.kt
```diff
@@ -67,7 +73,7 @@ public class SettingsFragment : PreferenceFragment(), Preference.OnPreferenceCha
 
     override fun onPreferenceClick(p0: Preference?): Boolean {
         if(AndroidAppSettings.KEY_ERASE_HISTORY.equals(p0?.key)) {
-            (getActivity() as SettingsActivity).clearHistory()
+            getSettingsActivity().clearHistory("History Erased")
```
ecgreb: We should add this to `strings.xml`.

app/src/main/kotlin/com/mapzen/erasermap/view/SettingsFragment.kt
```diff
@@ -106,6 +112,13 @@ public class SettingsFragment : PreferenceFragment(), Preference.OnPreferenceCha
         Tangram.setDebugFlag(DebugFlags.TANGRAM_INFOS, value)
     }
 
+    private fun updateCacheSearchResults(value: Boolean) {
+        settings?.isCacheSearchResultsEnabled = value
+        var status = if (value == true) "Enabled" else "Disabled"
```
ecgreb: `strings.xml`?

app/src/main/kotlin/com/mapzen/erasermap/view/MainActivity.kt
sarahlensing: I guess we don't actually need a preference listener, I had implemented it that way so that we only make changes when the value of 'cache search results changes'. We can definitely do something like what you're suggesting and rely on updating it when `onResume()` is called. Since we need to called `invalidateOptionsMenu()` for `onPrepareOptionsMenu()` to be called, I've made `onResume()` call `invalidateOptionsMenu()` and put the search bar configuration code in `onPrepareOptionsMenu()` since it is part of the menu. How does that sound?

app/src/main/kotlin/com/mapzen/erasermap/view/SettingsFragment.kt
sarahlensing: Fixed

app/src/main/kotlin/com/mapzen/erasermap/view/SettingsFragment.kt
sarahlensing: Fixed

app/src/main/kotlin/com/mapzen/erasermap/view/MainActivity.kt
```diff
@@ -182,6 +184,7 @@ public class MainActivity : AppCompatActivity(), MainViewController, RouteCallba
         app?.onActivityResume()
         autoCompleteAdapter?.clear()
         autoCompleteAdapter?.notifyDataSetChanged()
+        invalidateOptionsMenu()
```
ecgreb: Is this really required here? According to the docs for `onPrepareOptionsMenu()` it is called `called right before the menu is shown, every time it is shown.` Of course that doesn't mean this is the actual behavior ;)

https://developer.android.com/reference/android/app/Activity.html#onPrepareOptionsMenu(android.view.Menu)

On the other hand if `invalidateOptionsMenu()` is required to update whether we are caching search results I would think this needs to be called from `onResume()` since the activity is not destroyed but simply goes into the background when the settings activity is started.

ecgreb: Sorry - my mistake. The way this diff was displayed I thought this was in `onCreate()` when it in fact is in `onResume()`.

PR #447, opened 2016-03-02
Dont announce recalculating on reroute

### LINKED ISSUE #437: Don't announce recalculating

sarahlensing:
Closes #437


msmollin:
I'm OK with this. Feels like a concession but I can deal with it :D

PR #454, opened 2016-03-04
Send DNT header in tangram requests

### LINKED ISSUE #428: add DNT header to all API requests
This is how we know to drop the logs from Eraser map requests

sarahlensing:
Depends on: https://github.com/tangrams/tangram-es/pull/600

Closes #428


ecgreb:
Once the tangram change is deployed we should rebuild on circle using the "Rebuild without cache" option.

PR #462, opened 2016-03-08
438 clear mapdata

### LINKED ISSUE #438: Potrait<->Landscape messes up Client DataSource (specifically search pins/reverseGeo pins)
* **Device name:** Samsung Galaxy S5
* **Android Version:** 5.0
* **App build number:** master-1109
Client Datasources like search pins or reverseGeo/tappedPoi pins are not cleared after view is switched between landscape and portrait.
* **Steps to reproduce:**
    - start with portrait view 
    - tap on a poi to get a pin
    - switch to landscape view
    - tap on another poi
    - you should expect previous pin to be cleared and new pin to be added at the last poi tapped, but both pins are visible.
    - Similar thing will happen with search results.

screenshot_2016-03-01-15-27-14
screenshot_2016-03-01-15-27-23
screenshot_2016-03-01-15-27-29

sarahlensing:
- Moves searchResults & reverseGeocodeData MapData to MainPresenter so that objects persist on rotation changes
- Exposes MainPresenter's reverseGeo var so that it can be used to properly restore state on rotation changes
- Properly handles rotation changes when reverse geocoded results are shown (centers on map, updates options menu, etc)

Closes #438

### DIFF
--- a/app/src/main/kotlin/com/mapzen/erasermap/controller/MainActivity.kt
+++ b/app/src/main/kotlin/com/mapzen/erasermap/controller/MainActivity.kt
@@ -109,6 +109,8 @@ public class MainActivity : AppCompatActivity(), MainViewController, RouteCallba
     var autoCompleteAdapter: AutoCompleteAdapter? = null
     var optionsMenu: Menu? = null
     var findMe: MapData? = null
+    var reverseGeocodeData: MapData? = null
+    var searchResultsData: MapData? = null
     var startPin: MapData? = null
     var endPin: MapData? = null
     var poiTapPoint: FloatArray? = null
@@ -214,6 +216,8 @@ public class MainActivity : AppCompatActivity(), MainViewController, RouteCallba
         saveCurrentSearchTerm()
         routeModeView.clearRoute()
         findMe?.clear()
+        reverseGeocodeData?.clear()
+        searchResultsData?.clear()
         killNotifications()
     }
 
@@ -601,12 +605,12 @@ public class MainActivity : AppCompatActivity(), MainViewController, RouteCallba
 
         val properties = com.mapzen.tangram.Properties()
         properties.set(MAP_DATA_PROP_STATE, MAP_DATA_PROP_STATE_ACTIVE)
-        if (presenter?.reverseGeocodeData == null) {
-            presenter?.reverseGeocodeData = MapData("reverse_geocode")
-            Tangram.addDataSource(presenter?.reverseGeocodeData)
+        if (reverseGeocodeData == null) {
+            reverseGeocodeData = MapData("reverse_geocode")
+            Tangram.addDataSource(reverseGeocodeData)
         }
-        presenter?.reverseGeocodeData?.clear()
-        presenter?.reverseGeocodeData?.addPoint(properties, lngLat)
+        reverseGeocodeData?.clear()
+        reverseGeocodeData?.addPoint(properties, lngLat)
 
         mapController?.requestRender()
     }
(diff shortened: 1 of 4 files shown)


tallytalwar:
@sarahlensing Is this required now? "Moves searchResults & reverseGeocodeData MapData to MainPresenter so that objects persist on rotation changes", since `searchResults` features are already saved in the main presenter?

app/src/main/kotlin/com/mapzen/erasermap/presenter/MainPresenterImpl.kt
```diff
@@ -61,6 +64,7 @@ public open class MainPresenterImpl(val mapzenLocation: MapzenLocation, val bus:
     }
 
     override fun onReverseGeocodeResultsAvailable(searchResults: Result?) {
+        reverseGeo = true
```
tallytalwar: Not sure if `reverseGeo = true` is required here, since its already set before `onReverseGeocodeResultsAvailable` gets called on [line 411](https://github.com/mapzen/eraser-map/blob/438-clear-mapdata/app/src/main/kotlin/com/mapzen/erasermap/presenter/MainPresenterImpl.kt#L411) and [line422](https://github.com/mapzen/eraser-map/blob/438-clear-mapdata/app/src/main/kotlin/com/mapzen/erasermap/presenter/MainPresenterImpl.kt#L422).

sarahlensing:
@tallytalwar good question. it shouldn't be- let me test that clearing these MapData objects when the activity is destroyed works as expected

app/src/main/kotlin/com/mapzen/erasermap/presenter/MainPresenterImpl.kt
sarahlensing: You're right, its not required. Just tested that it works fine without it here and removed it.

app/src/main/kotlin/com/mapzen/erasermap/controller/MainActivity.kt
```diff
@@ -587,6 +585,10 @@ public class MainActivity : AppCompatActivity(), MainViewController, RouteCallba
     }
 
     override fun showReverseGeocodeFeature(features: List<Feature>) {
+        hideSearchResults()
+        centerOnCurrentFeature(features)
```
tallytalwar: Usually other mapping apps do not center the view on a reverse geo coded feature or a tapped poi feature. Do we need to center the map on these?

app/src/test/kotlin/com/mapzen/erasermap/controller/MainActivityTest.kt
```diff
@@ -276,7 +276,7 @@ public class MainActivityTest {
         features.add(getTestFeature())
         features.add(getTestFeature())
         activity.showSearchResults(features)
-        val shadowSearchResults = ShadowExtractor.extract(activity.searchResults) as ShadowMapData
+        val shadowSearchResults = ShadowExtractor.extract(activity.presenter?.searchResultsData) as ShadowMapData
```
tallytalwar: oopsi, test might fail.

app/src/main/kotlin/com/mapzen/erasermap/controller/MainActivity.kt
sarahlensing: I don't think we need to but it definitely feels nicer as a user when it centers. @mjcunningham want to weigh in?

mjcunningham: I think we shouldn't move the map automatically in this case. The user will have their finger pressed on one location and if the map moves while their finger is still on the screen then they will no longer be pressing in the same spot.

sarahlensing: Alrighty, I'll remove this automatic centering and only center when the device rotates. FYI though- the map doesnt center until after the user lifts their finger and we get a geocoded result to center on.

tallytalwar:
Looks good. Will merge.

PR #475, opened 2016-03-10
Use lateinit for ivars that are injected

sarahlensing:


ecgreb:
Wow nice find!
This is so much cleaner than making all the injected properties nullable.

sarahlensing:
thanks! yeah, its definitely a nice improvement

Issue #327, opened 2016-02-05
Update icon pressed states to use gray ripple effect


ecgreb:
Remove green pressed state assets. Apply material ripple effect to all buttons and icons in app.

PR #484, opened 2016-03-11
Add ripple effect to all buttons

### LINKED ISSUE #459: Route preview buttons need pressed state
"View List" and "Start Navigation" buttons no longer have a pressed state that visually responds to user touch.

sarahlensing:
Closes #459 #327


app/src/main/res/drawable/bg_black_gray_border.xml
```diff
@@ -1,12 +1,16 @@
 <?xml version="1.0" encoding="utf-8"?>
-<shape xmlns:android="http://schemas.android.com/apk/res/android"
-    android:shape="rectangle">
+<layer-list xmlns:android="http://schemas.android.com/apk/res/android">
```
ecgreb: If we use `<ripple>` instead of `<layer-list>` this appears to produce the desired effect for all the buttons and as an added bonus the tests no longer crash. For example:

```xml
<?xml version="1.0" encoding="utf-8"?>
<ripple xmlns:android="http://schemas.android.com/apk/res/android"
    android:color="?android:colorControlHighlight">
    <item>
        <shape xmlns:android="http://schemas.android.com/apk/res/android"
            android:shape="rectangle">

            <solid
                android:color="@color/color_primary" />

            <stroke
                android:width="1dp"
                android:color="@color/gray" />
        </shape>
    </item>
</ripple>
```

sarahlensing: You're right- looks like it does fix the tests! I had erroneously stayed away from `ripple` because i saw it introduced a requirement of minSdkVersion 21 however placing these in the qualified drawable folder doesn't require we bump the minSdkVersion.

PR #512, opened 2016-03-16
Adds settings menu "About" link

### LINKED ISSUE #495: Add terms/ privacy links

ecgreb:
This feature is unit tested using the new mockable Android jar framework.

This is a fairly simple test case but requires a decent amount of work on the test harness to get it able to run since this is the first time we are using the mockable Android jar.

I still don't like that that class under test `ViewAboutPreference` has to be mocked in order to successfully run the tests against it. Otherwise `getContext()` always returns null inside the `onClick()` method.

When mocking the class under test its important to call `Mockito.when(classUnderTest.methodUnderTest()).thenCallRealMethod()` to ensure the actual behavior is being tested and you are not just testing the mock.

Closes #495


mjcunningham:
lookin good :+1:

Issue #516, opened 2016-03-18
Crash: org.xmlpull.v1.XmlPullParserException


mjcunningham:
android.graphics.drawable.LayerDrawable.inflate

PR #517, opened 2016-03-18
Use colors for press states in layer-list

### LINKED ISSUE #515: Blocking crash on a Asus ZenFone 6 running Android 4.4.2

sarahlensing:
Since we are using `<ripple>` for api 21, simply use different colors for pressed states

Closes #515 #516


ecgreb:
Looks good. Tested on Galaxy S2 running Android 4.0.4.

PR #530, opened 2016-03-23
Move currentSnapLocation to RoutePresenter

sarahlensing:
- Persist currentSnapLocation across screen rotations
- Center map if device rotated in route mode
- Some Kotlin 1.0.1 cleanup

### DIFF
--- a/app/src/main/kotlin/com/mapzen/erasermap/presenter/RoutePresenter.kt
+++ b/app/src/main/kotlin/com/mapzen/erasermap/presenter/RoutePresenter.kt
@@ -6,26 +6,27 @@ import com.mapzen.erasermap.view.RouteViewController
 import com.mapzen.valhalla.Instruction
 import com.mapzen.valhalla.Route
 
-public interface RoutePresenter {
+interface RoutePresenter {
     companion object {
-        public val GESTURE_MIN_DELTA = 0.0001f
+        val GESTURE_MIN_DELTA = 0.0001f
     }
 
-    public var routeController: RouteViewController?
-    public var currentInstructionIndex: Int
+    var routeController: RouteViewController?
+    var currentInstructionIndex: Int
+    var currentSnapLocation: Location?
 
-    public fun onLocationChanged(location: Location)
-    public fun onRouteStart(route: Route?)
-    public fun onRouteResume(route: Route?)
-    public fun onMapPan(deltaX: Float, deltaY: Float)
-    public fun onResumeButtonClick()
-    public fun onInstructionPagerTouch()
-    public fun onInstructionSelected(instruction: Instruction)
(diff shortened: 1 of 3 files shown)


app/src/main/kotlin/com/mapzen/erasermap/view/RouteModeView.kt
```diff
@@ -431,6 +433,12 @@ class RouteModeView : LinearLayout, RouteViewController, ViewPager.OnPageChangeL
         initInstructionAdapter()
         this.visibility = View.VISIBLE
         routePresenter.onRouteResume(route)
+        Handler().postDelayed( {
+            if (routePresenter.currentSnapLocation != null) {
+                showRouteIcon(routePresenter.currentSnapLocation!!)
```
ecgreb: Introducing a read-only local variable might be a better way to express this without resorting to forced de-reference using `!!` operator:

```kotlin
        Handler().postDelayed( {
            val currentSnapLocation = routePresenter.currentSnapLocation
            if (currentSnapLocation != null) {
                showRouteIcon(currentSnapLocation)
            }
            centerMapOnCurrentLocation()
        }, 100)
```

app/src/main/kotlin/com/mapzen/erasermap/presenter/RoutePresenter.kt
```diff
@@ -6,26 +6,27 @@ import com.mapzen.erasermap.view.RouteViewController
...
-    public fun onInstructionPagerTouch()
-    public fun onInstructionSelected(instruction: Instruction)
-    public fun onUpdateSnapLocation(location: Location)
-    public fun onRouteClear()
-    public fun onMapListToggleClick(state: MapListToggleButton.MapListState)
-    public fun onRouteCancelButtonClick()
-    public fun mapZoomLevelForCenterMapOnLocation(location: Location): Float
-    public fun isTrackingCurrentLocation(): Boolean
-    public fun onSetCurrentInstruction(index: Int)
+    fun onLocationChanged(location: Location)
```
ecgreb: Without all the redundant modifiers these are so much more `fun` :)

app/src/main/kotlin/com/mapzen/erasermap/view/RouteModeView.kt
sarahlensing: Good point, let me update that now

sarahlensing: Fixed

app/src/main/kotlin/com/mapzen/erasermap/presenter/RoutePresenter.kt
sarahlensing: :P

PR #534, opened 2016-03-28
Use confidence score

### LINKED ISSUE #524: Use confidence score to influence reverse geocoding
Currently, the user taps the map, we send the coordinates to pelias, then drop the pin on the map where we get coordinates back from /reverse. This breaks down in areas with limited data coverage becuase the user will have their finger in one spot and we'll drop the pin in another. 

We can fix this by using the [confidence](https://mapzen.com/documentation/search/reverse/#confidence-scores-for-the-results) feature in pelias. 

We can drop the pin on the map where the user taps then send the coordinates to pelias. Based on the confidence score, we can alter the display text to say "Dropped pin _near_ feature XYZ" rather than snapping the pin to feature XYZ.

sarahlensing:
- Drop reversegeo pin where user presses
- Prefix label with "Near:" when confidence < 0.9
- Route to lat/lng instead of feature where confidence < 0.9
- Fix bug where unable to reverse geocode after coming back from results
- Add simple mechanism to version and clear cached search results

Depends on:
https://github.com/pelias/pelias-android-sdk/pull/40
https://github.com/pelias/pelias-android-sdk/pull/41

Closes #524


ecgreb:
Tapping a map feature that has an OSM ID results in a call to the Pelias `/place` endpoint. The response for this endpoint does not contain `confidence` resulting in "Near: ..." being displayed even when single tapping directly on the POI. For example:

https://pelias.mapzen.com/v1/place?ids=osm%3Avenue%3A360955274&api_key=search-BcgTWes

image

ecgreb:
Also minor note - find me button and attribution needs to be re-positioned higher on the screen when reverse geocoding (similar to normal search results).

sarahlensing:
Good catch, I updated the code to account for features without a confidence score. I also decreased the threshold from 0.9 to 0.8

I'll create a new ticket for the UI bug as discussed as it wasn't introduced in this ticket

app/src/test/kotlin/com/mapzen/erasermap/presenter/MainPresenterTest.kt
```diff
@@ -212,10 +212,10 @@ class MainPresenterTest {
         assertThat(mainController.isRoutePreviewVisible).isTrue()
     }
 
-    @Test fun onRoutePreviewEvent_shouldDisableReverseGeocode() {
```
ecgreb: I think we do still want to disable reverse geocode in route preview mode. Otherwise you can wind up with a situation with two destination pins like this which can be confusing.

image

app/src/main/kotlin/com/mapzen/erasermap/model/AndroidAppSettings.kt
```diff
@@ -1,30 +1,37 @@
 package com.mapzen.erasermap.model
 
+import android.content.Context
 import android.location.Location
 import android.preference.PreferenceManager
+import android.util.Log
```
ecgreb: Unused imports `Log` and `Toast`

PR #535, opened 2016-03-28
Focus & unfocus audio for ducking

### LINKED ISSUE #336: Add audio ducking capability
* **Android Version:** 5.0
* **App build number:** master-792
* **What did you expected to happen?** When using my phone for listening to music and navigation for a trip - I expect that the music volume is lowered when a navigation command is given and then the music volume returned to same level after the guidance command. This is the audio ducking capability that other nav apps implement.
* **What happened instead?** The music volume remained constant and the navigation command could not be heard.
* **Steps to reproduce:** Play music on phone while using eraser map in nav mode.

sarahlensing:
Depends on releasing `1.4.0` from https://github.com/mapzen/speakerbox/pull/9/files

Closes #336


ecgreb:
We should really consider implementing audio ducking only when a verbal instruction is currently being spoken rather then requesting audio focus for the entire duration of a route.

dgearhart:
Agreed @ecgreb - that is how audio ducking is typically implemented 
Has this been pushed yet? I tested yesterday and could not hear the verbal commands

ecgreb:
Not yet. TTS work happening [here](https://github.com/mapzen/speakerbox/pull/12) and [here](https://github.com/mapzen/speakerbox/pull/13).

dgearhart:
Thanks @ecgreb

PR #539, opened 2016-04-01
Make directions list consistent

### LINKED ISSUE #500: Add Current Location as first item in direction list view
Currently the directions list before starting a route has it and not the list view during navigation. Will be good to make them consistent

### LINKED ISSUE #501: Directions list view styling consistency
Currently the "view List" before starting a route and after are styled differently. They can look pretty similar

*View List before starting navigation*
(No map toggle icon, no cancel icon)
Animation: List can slide up and the mode icons row can collapse
Info: Distance + Time Remaining

*View List during navigation*
(Add 'current location https://github.com/mapzen/eraser-map/issues/500 and remove cancel icon)
Info: Distance + Time Remaining

sarahlensing:
- Remove use of `InstructionListActivity` and animate `DirectionListView` instead
- Always show current location in `DirectionsListAdapter`

Closes #501 
Closes #500


ecgreb:
In routing mode current instruction highlight is now off by 1. Also clicking on an instruction in the list zooms to the instruction 1 before that which was clicked. Should be easily fixed by updating `getView(...)` and `onDirectionClicked(Int)` in `DirectionListAdapter` to account for `CURRENT_INSTRUCTION_OFFSET`.

There are only two hard problems in computer science...

app/src/main/kotlin/com/mapzen/erasermap/view/RouteModeView.kt
```diff
@@ -50,9 +50,6 @@ class RouteModeView : LinearLayout, RouteViewController, ViewPager.OnPageChangeL
     val mapListToggle: MapListToggleButton by lazy {
         findViewById(R.id.map_list_toggle)as MapListToggleButton
     }
-    val routeCancelButton: ImageView by lazy {
```
ecgreb: Why was the route cancel button removed? I think it might have been left out of the mockups accidentally ... @souperneon can you confirm?

app/src/main/res/layout/view_distance_title.xml
```diff
@@ -0,0 +1,68 @@
...
+              android:ellipsize="end"
+              android:fontFamily="sans-serif-condensed"
+              android:maxLines="1"
+              android:singleLine="true"
+              android:textColor="@color/gray"
+              android:textSize="@dimen/font_medium_2"
+              android:textStyle="bold" />
+      </LinearLayout>
+
+</RelativeLayout>
```
ecgreb: New line

app/src/main/kotlin/com/mapzen/erasermap/view/InstructionListActivity.kt
```diff
@@ -1,59 +0,0 @@
...
-        @JvmStatic val EXTRA_STRINGS = "instruction_strings"
-        @JvmStatic val EXTRA_TYPES = "instruction_types"
-        @JvmStatic val EXTRA_DISTANCES = "instruction_distances"
-        @JvmStatic val EXTRA_REVERSE = "reverse"
-        @JvmStatic val EXTRA_DESTINATION = "destination"
-    }
-
-    override fun onCreate(savedInstanceState: Bundle?) {
-        super.onCreate(savedInstanceState)
-        setContentView(R.layout.activity_instructions)
```
ecgreb: `activity_instructions.xml` can be safely removed as well.

app/src/main/kotlin/com/mapzen/erasermap/view/RouteModeView.kt
souperneon: @sarahlensing and I discussed this some. Currently, 'x' in the nav map view exits nav and list view just collapses list. We both agreed that exiting from list view is not ideal for users and having the 'x' do different things on different screens is not good either.
So… the 'x' is no more available on the list view screen. We noticed that Google doesn't have it either. What are your thoughts @ecgreb?

ecgreb: Ok that makes sense. However the X is no longer available on the map view either.

souperneon: wait really? why not?

ecgreb: Seems like it was just a mixup. We will be adding it back on the map view. Thanks for the clarification!

sarahlensing: @souperneon removing it from both places was what we had discussed and what is in your mocks. let me show you what this looks like on the device to hopefully remove confusion though!

ecgreb: Per discussion with @souperneon IRL I'm going to restore route cancel button (map view only) and it will be behave same as the system back button.

PR #547, opened 2016-04-15
13 stylesheet names

sarahlensing:
Depends on https://github.com/tangrams/bubble-wrap/pull/176


ecgreb:
Merged via https://github.com/mapzen/eraser-map/pull/553

PR #559, opened 2016-05-10
550 search results editable listview

### LINKED ISSUE #550: Search term should not be editable in results list view

sarahlensing:
Closes #550

### DIFF
--- a/app/src/main/kotlin/com/mapzen/erasermap/controller/MainActivity.kt
+++ b/app/src/main/kotlin/com/mapzen/erasermap/controller/MainActivity.kt
@@ -428,6 +428,7 @@ class MainActivity : AppCompatActivity(), MainViewController, RouteCallback,
                 expandSearchView()
             } else if (presenter.resultListVisible) {
                     onCloseAllSearchResults()
+                    enableSearch()
                 } else {
                 searchView?.setQuery(presenter.currentSearchTerm, false)
             }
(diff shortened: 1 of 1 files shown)


app/src/main/kotlin/com/mapzen/erasermap/controller/MainActivity.kt
```diff
@@ -436,6 +438,13 @@ class MainActivity : AppCompatActivity(), MainViewController, RouteCallback,
         val cacheSearches = PreferenceManager.getDefaultSharedPreferences(this)
                 .getBoolean(AndroidAppSettings.KEY_CACHE_SEARCH_HISTORY, true)
         searchView?.setCacheSearchResults(cacheSearches)
+        val closeButton = searchView?.findViewById(R.id.search_close_btn) as ImageView
+        closeButton.setOnClickListener {
```
ecgreb: Nice job.

Do you think it would be possible to either: 1. Hide the close button when viewing the search results list or 2. Override the behavior to clear the current set of results? Thoughts about which would be better?

It just feels kind of broken right now that you can see the close button but it doesn't do anything when clicked in this mode.

sarahlensing: I like the first suggestion to hide the close button, I think it feels most natural. Good call, I like it better than overriding the click listener!

app/src/main/kotlin/com/mapzen/erasermap/controller/MainActivity.kt
```diff
@@ -1095,6 +1100,18 @@ class MainActivity : AppCompatActivity(), MainViewController, RouteCallback,
         animations.start()
     }
 
+    fun enableSearch() {
+        searchView?.inputType = 1
```
ecgreb: Changes look good. The only other thing I see is maybe we should use the constant values `InputType.TYPE_CLASS_TEXT` (1) here and `InputType.TYPE_NULL` (0) below for clarity?

sarahlensing: Good call. Fixed.

PR #580, opened 2016-05-20
Handle implicit intents

### LINKED ISSUE #545: Handle implicit intents
Both geo and map intents.

ecgreb:
Handle implicit geo and map intents generated by other applications.

* Updates AndroidManifest.xml to receive geo and maps intents
* Log incoming intent and forward to MainActivity
* Perform input validation and pass valid query to parser
* Adds logic to parse query string
* Populate current search term with IntentQueryParser result
* Execute search on incoming query string
* Set map position prior to search based on focus point if available
* Handle implicit intent even if app is already running in background

Closes #545

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -32,8 +32,26 @@
             android:screenOrientation="portrait">
             <intent-filter>
                 <action android:name="android.intent.action.MAIN" />
-
                 <category android:name="android.intent.category.LAUNCHER" />
+            </intent-filter>
+            <intent-filter>
+                <action android:name="android.intent.action.VIEW" />
+                <data android:scheme="geo" />
+                <category android:name="android.intent.category.DEFAULT" />
+            </intent-filter>
+            <intent-filter>
+                <action android:name="android.intent.action.VIEW" />
+                <data
+                    android:scheme="http"
+                    android:host="maps.google.com" />
+                <category android:name="android.intent.category.DEFAULT" />
+            </intent-filter>
+            <intent-filter>
+                <action android:name="android.intent.action.VIEW" />
+                <data
+                    android:scheme="https"
+                    android:host="maps.google.com" />
+                <category android:name="android.intent.category.DEFAULT" />
             </intent-filter>
         </activity>
         <receiver android:name=".receiver.MockLocationReceiver">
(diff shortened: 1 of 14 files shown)


sarahlensing:
Looks good! A couple issues I'm seeing- it looks like we arent decoding the query and it also seems like we aren't executing a search if the app has already been launched and we receive an implicit intent. It could be because the MainActivity is not setup to handle intents, only the InitActivity is. See below:
screenshot_20160520-154752
screenshot_20160520-154819

ecgreb:
`Uri#getQuery()` _claims_ to decode the query string but that doesn't mean I trust it. What does the intent data look like when `InitActivity` logs it to the console? Also what app are you using to generate the intents?

https://developer.android.com/reference/android/net/Uri.html#getQuery()

I am invoking `presenter.onIntentQueryReceived(intent?.data?.query)` from `MainActivity#onNewIntent(Intent)` specifically to handle the case where the app is already running in the background. I wonder if this is another issue specific to Samsung devices (similar to vanishing API keys) because of the way TouchWiz handles background apps. On the Robin implicit intents are handled correctly when app is already open.

sarahlensing:
When it's logged its `Incoming implicit intent: geo:0,0?q=Pier+97` however `data.toString()` is `geo:0,0?q=Pier+97` :/

I'm using Google Calendar to generate the intent.

It could definitely be an issue related to that. As a simple test, I set a breakpoint in `onNewIntent` and found it was never called. Then I added the intent filters to `MainActivity` and rebuilt and saw that `onNewIntent` in `MainActivity` was invoked however this breaks the case where the app isn't running because the ApiKeys are initialized/configured in InitActivity. There might be another way to handle this but adding the intent filters to `MainActivity` seems to have fixed what I saw. If we initialize the ApiKeys in the init block of `ApiKeys` and remove it from `InitActivity` it would fix both issues. Of course, we need to make sure this still works on the Robin...

ecgreb:
Thanks for the additional info and debugging. Its very strange indeed that `onNewIntent(...)` doesn't even get called.

ecgreb:
@sarahlensing both issues have been addressed in subsequent commits. Thanks for your help tracking them down!

sarahlensing:
@ecgreb The changes look good- merging!

PR #595, opened 2016-05-25
Move ApiKey configuration into init block

sarahlensing:
- `MainActivity` will initialize the keys when it is created
- Initialization will happen once per application lifecycle


ecgreb:
LGTM 👍 - however there are some merge conflicts that need rebased.

PR #627, opened 2016-06-09
Deactivate find me tacking upon entering search results or route preview

### LINKED ISSUE #623: Deactivate "find me" when in search or routing mode
When find me is on and you're trying to do something else it will keep snapping back to your current location.

ecgreb:
Closes #623


sarahlensing:
This looks good. I had one thought though- do you think we should re-enable tracking when the user exits routing/search results if it was on when they began either of these? Or is it reasonable to leave it disabled?

mjcunningham:
I think we should leave it disabled. It's weird to have to pop you back to current location without you actively hitting the find me button.

sarahlensing:
Sounds good!

PR #646, opened 2016-06-20
Fix mock location

### LINKED ISSUE #584: Unable to set mock location
The "find me" button no longer recognizes mock locations set in the debug menu since migrating to the Mapzen SDK. Same issue with mock route gpx replay.

sarahlensing:
- Because the sdk's `LostApiClient is disconnected when starting navigation/pausing the main activity we need to re-enable mock mode when `mapzenMap?.isMyLocationEnabled = true` because a new `FusedLocationProviderImpl` is created (and by default has a `FusionEngine`, not a `MockEngine`)
- When starting navigation, we need to set the trace file after the location client has been connected and before we request location updates
- We need to set the mock location in the `MockLocationReceiver` so that the sdk will handle updating the current location tracker properly

Depends on https://github.com/mapzen/LOST/pull/54/files

Closes #584

### DIFF
--- a/app/src/main/kotlin/com/mapzen/erasermap/model/MapzenLocationImpl.kt
+++ b/app/src/main/kotlin/com/mapzen/erasermap/model/MapzenLocationImpl.kt
@@ -16,8 +16,7 @@ import com.mapzen.pelias.BoundingBox
 import com.squareup.otto.Bus
 import com.squareup.otto.Subscribe
 
-public class MapzenLocationImpl(val locationClient: LostApiClient,
-        val settings: AppSettings,
+public class MapzenLocationImpl(val settings: AppSettings,
         val bus: Bus,
         val application: EraserMapApplication,
         val permissionManager: PermissionManager) : MapzenLocation {
@@ -32,12 +31,13 @@ public class MapzenLocationImpl(val locationClient: LostApiClient,
     }
 
     override var mapzenMap: MapzenMap? = null
+    override var locationClient: LostApiClient? = null
 
     private var previousLocation: Location? = null
 
     private fun connect() {
-        if (!locationClient.isConnected) {
-            locationClient.connect()
+        if (locationClient != null && !locationClient!!.isConnected) {
+            locationClient?.connect()
 
             if (settings.isMockLocationEnabled) {
                 LocationServices.FusedLocationApi?.setMockMode(true)
@@ -47,7 +47,9 @@ public class MapzenLocationImpl(val locationClient: LostApiClient,
     }
 
     private fun disconnect() {
-        locationClient.disconnect()
+        if (locationClient?.numberOfListeners() == 1) {
+            locationClient?.disconnect()
+        }
     }
 
     override fun getLastLocation(): Location? {
--- a/app/src/main/kotlin/com/mapzen/erasermap/model/MapzenLocationImpl.kt
+++ b/app/src/main/kotlin/com/mapzen/erasermap/model/MapzenLocationImpl.kt
@@ -16,7 +16,8 @@ import com.mapzen.pelias.BoundingBox
 import com.squareup.otto.Bus
 import com.squareup.otto.Subscribe
 
-public class MapzenLocationImpl(val settings: AppSettings,
+public class MapzenLocationImpl(val locationClient: LostApiClient,
+        val settings: AppSettings,
         val bus: Bus,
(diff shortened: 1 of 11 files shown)


app/src/main/kotlin/com/mapzen/erasermap/model/MapzenLocationImpl.kt
```diff
@@ -38,11 +36,11 @@ public class MapzenLocationImpl(val locationClient: LostApiClient,
...
-
-            if (settings.isMockLocationEnabled) {
-                LocationServices.FusedLocationApi?.setMockMode(true)
-                LocationServices.FusedLocationApi?.setMockLocation(settings.mockLocation)
-            }
+        }
+        if (settings.isMockLocationEnabled) {
+            LocationServices.FusedLocationApi?.setMockMode(true)
+            LocationServices.FusedLocationApi?.setMockLocation(settings.mockLocation)
+            LocationServices.FusedLocationApi?.setMockTrace(settings.mockRoute)
```
ecgreb: There is a separate checkbox preference for mock routing. We should check `settings.isMockRouteEnabled` before setting the trace file.

sarahlensing: Nice catch. I've pushed changes to check this setting. I also updated the way our settings behave so that enabling mock routing enables mock location as well.

app/src/main/kotlin/com/mapzen/erasermap/view/SettingsFragment.kt
```diff
@@ -1,6 +1,8 @@
 package com.mapzen.erasermap.view
 
+import android.content.SharedPreferences
```
ecgreb: Unused import

app/src/main/kotlin/com/mapzen/erasermap/view/SettingsFragment.kt
```diff
@@ -118,6 +134,24 @@ public class SettingsFragment : PreferenceFragment(), Preference.OnPreferenceCha
         getSettingsActivity().clearHistory(title)
     }
 
+    private fun updateMockTracePref(mockLocationEnabled: Boolean) {
+        if (!mockLocationEnabled) {
+            val prefs = PreferenceManager.getDefaultSharedPreferences(this.context)
```
ecgreb: `Fragment.getContext()` requires API 23. This will crash on older versions of Android

https://developer.android.com/reference/android/app/Fragment.html#getContext()

app/src/main/kotlin/com/mapzen/erasermap/view/SettingsFragment.kt
```diff
@@ -118,6 +134,24 @@ public class SettingsFragment : PreferenceFragment(), Preference.OnPreferenceCha
...
+            prefs.edit().putBoolean(AndroidAppSettings.KEY_MOCK_ROUTE_ENABLED, false).commit()
+            val pref = findPreference(AndroidAppSettings.KEY_MOCK_ROUTE_ENABLED) as CheckBoxPreference
+            pref.isChecked = false
+        }
+    }
+
+    private fun updateMockLocationPref(mockTraceEnabled: Boolean) {
+        if (mockTraceEnabled) {
+            val prefs = PreferenceManager.getDefaultSharedPreferences(this.context)
```
ecgreb: Fragment.getContext() requires API 23. This will crash on older versions of Android

https://developer.android.com/reference/android/app/Fragment.html#getContext()

app/src/main/kotlin/com/mapzen/erasermap/view/SettingsFragment.kt
sarahlensing: Rm'd

app/src/main/kotlin/com/mapzen/erasermap/view/SettingsFragment.kt
sarahlensing: Updated to access context through activity.

app/src/main/kotlin/com/mapzen/erasermap/view/SettingsFragment.kt
sarahlensing: Updated to access context through activity.

ecgreb:
So glad to have this feature back!

PR #651, opened 2016-06-21
Hide resume button on exit routing mode

### LINKED ISSUE #602: Routing mode uses incorrect camera (sometimes)

ecgreb:
Explicitly hide the resume button when exiting routing mode. This ensures the correct camera and map position is set when starting a new route.

Fixes #602

### DIFF
--- a/app/src/main/kotlin/com/mapzen/erasermap/controller/MainActivity.kt
+++ b/app/src/main/kotlin/com/mapzen/erasermap/controller/MainActivity.kt
@@ -1176,6 +1176,7 @@ class MainActivity : AppCompatActivity(), MainViewController,
         supportActionBar?.hide()
         routeModeView.route = null
         routeModeView.hideRouteIcon()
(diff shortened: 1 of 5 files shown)


sarahlensing:
This is a good addition but I can still see the problem (to consistently reproduce it I touch and hold my press on the map and then hit 'start navigation'). What if we added a flag in `MainPresenter#clickStartNavigation` called `startingRoute` and set it to be true at the beginning of the method/false at the end and then in `MainActivity#setRoutingCamera` always set the routing camera if we are `startingRoute`. That way on reroutes we would still check that the resume button was hidden and starting a route would always set the camera

ecgreb:
Good catch! Will push an update.

ecgreb:
So it turns out if you interacted with the map on the route preview screen in any way (pan, zoom, rotate, etc) then it would "show" the resume button and the camera would not be set.

This new commit manually re-hides the resume button when a new route is started.

PR #657, opened 2016-06-23
color matching and two sizes

souperneon:


ecgreb:
Looking good!

Dash `-` is not allowed in resource file names. These need to replaced with underscore `_`.

Also I don't think we need large maneuver icons for transit since it does not have live navigation view? (list view only)

@souperneon do you want to make these changes or should I?

souperneon:
I trust you @ecgreb :)

PR #660, opened 2016-07-06
multi-modal route preview design

sarahlensing:
- Adds `MultiModalDirectionListAdapter` for handling display of multimodal preview
- Adds `InstructionGrouper` to handle breaking instructions into `InstructionGroup` objects
- Adds `ListRowItem` for use in `MultiModalDirectionListAdapter`, links `InstructionGroup` to view display properties
- Removes `TravelMode`, `TravelType` code from `MultiModalHelper` (read from on-the-road)

- Future PRs will update the map with colored routing polyline/station icons and will add proper icons

Closes #660


app/src/main/kotlin/com/mapzen/erasermap/model/InstructionGroup.kt
```diff
@@ -0,0 +1,63 @@
...
+    val instructions: ArrayList<Instruction>) {
+
+  val totalDistance: Int by lazy { calculateTotalDistance() }
+  val totalTime: Int by lazy { calculateTotalTime() }
+  var firstStationName: String? = null
+  var numberOfStops: String? = null
+
+  private fun calculateTotalDistance(): Int {
+    var total = 0
+    for(i in 0..instructions.size - 1) {
```
msmollin: Why the -1's here? is Size 1-indexed or something?

app/src/main/kotlin/com/mapzen/erasermap/view/MultiModalDirectionListAdapter.kt
```diff
@@ -0,0 +1,423 @@
...
+import java.util.ArrayList
+import java.util.HashMap
+import java.util.HashSet
+import java.util.regex.Pattern
+
+class MultiModalDirectionListAdapter(val context: Context, val instructionGrouper: InstructionGrouper,
+    val reverse : Boolean?, val multiModalHelper: MultiModalHelper) : BaseAdapter() {
+
+  companion object {
+    const val SEC_PER_MIN = 60
```
msmollin: I feel like there's another class already available with SEC_PER_MIN - am I imagining that?

app/src/main/kotlin/com/mapzen/erasermap/view/MultiModalDirectionListAdapter.kt
```diff
@@ -0,0 +1,423 @@
...
+        titleView.text = instruction.getHumanTurnInstruction()
+        distanceView.distanceInMeters = distance
+
+        holder.instructionsContainer.addView(instructionRow)
+      }
+    }
+
+  }
+
+  private fun setTransitRow(position: Int, holder: TransitViewHolder) {
```
msmollin: This function seems to do a lot. How hard would it be to refactor it into some smaller utility functions? Maybe a couple getters for some of the model related things?

app/src/main/kotlin/com/mapzen/erasermap/view/MultiModalDirectionListAdapter.kt
```diff
@@ -0,0 +1,423 @@
...
+          as DistanceTimeExpanderLayout
+      val textView = view.findViewById(R.id.distance_time_text_view) as TextView
+      val builder = StringBuilder()
+      builder.append(context.getString(R.string.comma))
+      builder.append(" ")
+      textView.text = builder.toString()
+    }
+  }
+
+  class TransitViewHolder(view: View, context: Context): ViewHolder() {
```
msmollin: Note entirely following the utility of this class and the one above it. Are they just convenience or do they actually have some bearing on the UI?

app/src/main/kotlin/com/mapzen/erasermap/view/ViewRecycler.kt
```diff
@@ -0,0 +1,35 @@
+package com.mapzen.erasermap.view
+
+import android.view.View
+import java.util.HashMap
+import java.util.HashSet
+
+class ViewRecycler {
```
msmollin: This seems a little spartan for a view recycler, at least compared to android's RecyclerView and iOS's myriad of recycling view components. This seems to be more like a singleton view holder so you never have more than one type of view, or am I reading it wrong?

app/src/test/kotlin/com/mapzen/erasermap/model/InstructionGrouperTest.kt
```diff
@@ -0,0 +1,49 @@
...
+    assertThat(grouper.numInstructionsInGroup(0)).isEqualTo(3)
+    assertThat(grouper.numInstructionsInGroup(1)).isEqualTo(1)
+    assertThat(grouper.numInstructionsInGroup(2)).isEqualTo(1)
+    assertThat(grouper.numInstructionsInGroup(3)).isEqualTo(3)
+  }
+
+  @Test fun getInstructionGroup_shouldReturnCorrectGroup() {
+    assertThat(grouper.getInstructionGroup(0).instructions[0].getHumanTurnInstruction()).isEqualTo(
+        "Instruction 0")
+    assertThat(grouper.getInstructionGroup(1).instructions[0].getHumanTurnInstruction()).isEqualTo(
```
msmollin: Seems Instruction 3 is mentioned twice here. Intended?

app/src/main/kotlin/com/mapzen/erasermap/model/InstructionGroup.kt
msmollin: Looking through the rest of the PR it seems like there's an indexing issue here that's somewhat systemic. Having to constantly `-1` the size everywhere seems odd.

sarahlensing: It is indexed but looking at the documentation, I see I can use the `instructions.indices` instead. That is indeed better than what I've currently got. Updated.

app/src/main/kotlin/com/mapzen/erasermap/view/MultiModalDirectionListAdapter.kt
sarahlensing: `TimeUnit.SECONDS.toMinutes()`! 
Updated.

app/src/main/kotlin/com/mapzen/erasermap/view/MultiModalDirectionListAdapter.kt
sarahlensing: Good call, refactored into smaller functions and moved some code into the model.

app/src/main/kotlin/com/mapzen/erasermap/view/MultiModalDirectionListAdapter.kt
sarahlensing: They follow the "view holder" pattern and allow access to the underlying view directly, without having to call `findViewByID` everytime they are accessed

app/src/main/kotlin/com/mapzen/erasermap/view/ViewRecycler.kt
sarahlensing: This definitely is spartan but it allows recycling the views of the inner `LinearLayout` which doesn't have recycling built into it. There are two types of views that get recycled- `R.layout.instruction_row` and `R.layout.transit_station_row`

app/src/test/kotlin/com/mapzen/erasermap/model/InstructionGrouperTest.kt
sarahlensing: Yes, intended. I add two of that instruction into `instructions`

msmollin: Ah so you did.

app/src/main/kotlin/com/mapzen/erasermap/view/DynamicChildHeightRelativeLayout.kt
```diff
@@ -0,0 +1,44 @@
...
+
+import android.content.Context
+import android.util.AttributeSet
+import android.view.View
+import android.widget.RelativeLayout
+import java.util.ArrayList
+
+abstract class DynamicChildHeightRelativeLayout : RelativeLayout {
+
+  constructor(context: Context) : super(context) {}
```
ecgreb: Interestingly enough it seems the curly braces `{}` can be safely deleted in Kotlin if the constructor body is empty and the app still compiles and runs fine ;)

app/src/main/kotlin/com/mapzen/erasermap/view/DynamicChildHeightRelativeLayout.kt
```diff
@@ -0,0 +1,44 @@
...
+    for (i in 0..idsForDynamicChildren().size - 1) {
+      val viewId = idsForDynamicChildren()[i]
+      (views as ArrayList<View>).add(findViewById(viewId))
+    }
+  }
+
+  override fun onLayout(changed: Boolean, l: Int, t: Int, r: Int, b: Int) {
+    super.onLayout(changed, l, t, r, b)
+    val height = Math.abs(b - t)
+    for (i in 0..views!!.size - 1) {
```
ecgreb: If `views` is `null` this will crash. Add a null check before the loop?

app/src/main/kotlin/com/mapzen/erasermap/view/DynamicChildHeightRelativeLayout.kt
```diff
@@ -0,0 +1,44 @@
...
+    super.onFinishInflate()
+    views = ArrayList<View>()
+    for (i in 0..idsForDynamicChildren().size - 1) {
+      val viewId = idsForDynamicChildren()[i]
+      (views as ArrayList<View>).add(findViewById(viewId))
+    }
+  }
+
+  override fun onLayout(changed: Boolean, l: Int, t: Int, r: Int, b: Int) {
+    super.onLayout(changed, l, t, r, b)
```
ecgreb: Calling `super.onLayout(...)` here will cause the dynamic children to be laid out twice which is not super performant. Perhaps we could avoid with something like this-- borrowed from `RelaiveLayout.onLayout(...)`.

```kotlin
  override fun onLayout(changed: Boolean, l: Int, t: Int, r: Int, b: Int) {
    val dynamicViews = views
    if (dynamicViews == null) {
      super.onLayout(changed, l, t, r, b)
      return
    }

    val count = childCount
    val height = Math.abs(b - t)

    for (i in 0..count - 1) {
      val child = getChildAt(i)

      if (child.visibility != GONE) {          
        val params = child.layoutParams as RelativeLayout.LayoutParams
        if (child in dynamicViews) {
          layoutView(child, b + height)
        } else {
          child.layout(params.mLeft, params.mRight, params.mTop, params.mBottom)
        }
      }
    }
  }
```

The only catch here is that `mLeft`, `mRight`, `mTop`, `mBottom` are private members of `RelativeLayout.LayoutParams` 😞  but maybe we can find a way to workaround that. If not I think its fine to keep as is for now and just keep an eye out for jank when scrolling the list.

app/src/main/kotlin/com/mapzen/erasermap/controller/MainActivity.kt
```diff
@@ -1074,9 +1084,25 @@ class MainActivity : AppCompatActivity(), MainViewController,
         routeBtmContainer.visibility = View.VISIBLE
         distanceView.distanceInMeters = routeManager.route?.getTotalDistance() as Int
         destinationNameTextView.text = simpleFeature.name()
-        previewDirectionListView.adapter = DirectionListAdapter(this, instructionStrings,
-                instructionTypes, instructionDistances, routeManager.reverse,
-                MultiModalHelper(routeManager.route?.rawRoute))
+        if (routeManager.type == Router.Type.MULTIMODAL) {
```
ecgreb: Next time we update the route preview logic we should look into moving the code to show/hide the instruction list into `RoutePreviewView` if possible.

app/src/main/kotlin/com/mapzen/erasermap/view/MultiModalDirectionListAdapter.kt
```diff
@@ -0,0 +1,423 @@
+package com.mapzen.erasermap.view
+
+import android.content.Context
+import android.graphics.Color
```
ecgreb: Unused imports

```kotlin
import android.graphics.Color
import android.graphics.Typeface
import android.text.SpannableString
import android.text.style.StyleSpan
import com.mapzen.valhalla.TransitInfo
import java.util.regex.Pattern
```

app/src/main/kotlin/com/mapzen/erasermap/model/MultiModalHelper.kt
```diff
@@ -1,67 +1,12 @@
 package com.mapzen.erasermap.model
 
 import com.mapzen.erasermap.R
+import com.mapzen.valhalla.Instruction
```
ecgreb: Unused import

app/src/main/kotlin/com/mapzen/erasermap/model/MultiModalHelper.kt
```diff
@@ -1,67 +1,12 @@
...
+import com.mapzen.valhalla.TravelType
 import org.json.JSONObject
 
 class MultiModalHelper(private val rawRoute: JSONObject?) {
```
ecgreb: Constructor param `rawRoute` is no longer used and can be safely removed.

app/src/main/res/layout/padding_section.xml
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" encoding="utf-8"?>
+<LinearLayout xmlns:android="http://schemas.android.com/apk/res/android"
```
ecgreb: Where is this layout used?

app/src/main/kotlin/com/mapzen/erasermap/view/PedestrianRowRelativeLayout.kt
```diff
@@ -0,0 +1,23 @@
+package com.mapzen.erasermap.view
+
+import android.content.Context
+import android.util.AttributeSet
+import android.view.View
```
ecgreb: Unused imports

```kotlin
import android.view.View
import android.widget.RelativeLayout
```

app/src/main/kotlin/com/mapzen/erasermap/view/DynamicChildHeightRelativeLayout.kt
sarahlensing: Removed

app/src/main/kotlin/com/mapzen/erasermap/view/DynamicChildHeightRelativeLayout.kt
sarahlensing: Added

app/src/main/kotlin/com/mapzen/erasermap/view/DynamicChildHeightRelativeLayout.kt
sarahlensing: All I can think of is using reflection or copying the entirety of `RelativeLayout` into this class. I think I'll leave it as is for now and keep an eye out as you suggest

app/src/main/kotlin/com/mapzen/erasermap/model/InstructionGroup.kt
```diff
@@ -0,0 +1,63 @@
+package com.mapzen.erasermap.model
```
sarahlensing: Good point, I've moved them

app/src/main/kotlin/com/mapzen/erasermap/controller/MainActivity.kt
sarahlensing: Moved!

app/src/main/kotlin/com/mapzen/erasermap/view/MultiModalDirectionListAdapter.kt
sarahlensing: Rm'd

app/src/main/kotlin/com/mapzen/erasermap/model/MultiModalHelper.kt
sarahlensing: Rm'd

app/src/main/kotlin/com/mapzen/erasermap/model/MultiModalHelper.kt
sarahlensing: Rm'd

app/src/main/res/layout/padding_section.xml
sarahlensing: Oops, its not used. Rm'd

app/src/main/kotlin/com/mapzen/erasermap/view/PedestrianRowRelativeLayout.kt
sarahlensing: Rm'd

PR #675, opened 2016-07-13
Return map's position for autocomplete query

### LINKED ISSUE #603: Use current map location when fetching autocomplete results
Currently uses raw device location

sarahlensing:
Closes #603


ecgreb:
Nice clean solution

PR #692, opened 2016-07-20
Fix incorrect results showing for reverse geocode

### LINKED ISSUE #672: Incorrect region text when tapping on city
When the map is zoomed out tapping on a city shows inconsistent results for region text leading to sometimes comical results. See examples below.

### I didn't realize Philly was part of NJ now.

image

### If Philly is part of NJ then of course Trenton is part of NY.

image

### Why just coordinates here? I think we can do better. Also no city labels on the map?

image

### And the best one yet submitted by one of our users... display bug or political satire?

image

sarahlensing:
- Pass poiTapPoint(1) for screenY, not poiTapPoint(0)
- Replace use of `android.graphics.PointF` in `TestMainController` with `com.mapzen.tangram.LngLat`
- Reformat `MainPresenterImpl`

Closes #672

### DIFF
--- a/app/src/main/kotlin/com/mapzen/erasermap/controller/MainActivity.kt
+++ b/app/src/main/kotlin/com/mapzen/erasermap/controller/MainActivity.kt
@@ -88,8 +88,6 @@ class MainActivity : AppCompatActivity(), MainViewController,
         SearchViewController.OnSearchResultSelectedListener {
 
     companion object {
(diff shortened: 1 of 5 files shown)


ecgreb:
Nice catch.

👍 for reformatting but in the future we should probably keep reformatting separate from actual code changes for the sake of code review and commit history.

PR #719, opened 2016-08-08
Move dagger compiler to apt scope and rm unused deps

sarahlensing:


app/build.gradle
```diff
@@ -146,4 +145,5 @@ dependencies {
   testCompile 'org.assertj:assertj-core:1.7.0'
   testCompile "org.jetbrains.kotlin:kotlin-reflect:$kotlin_version"
   testCompile 'org.json:json:20160212'
+  testCompile 'com.google.dagger:dagger-compiler:2.0'
```
ecgreb: Does `dagger-compiler` need to be included as both an `apt` and `testCompile` dependency? Is this what was causing https://github.com/mapzen/eraser-map/pull/713 to fail on CI?

If so we should extract a variable `$dagger_version` similar to `$kotlin_version` so we only need to update in one place.

sarahlensing: Good call. I updated the build script.

Yes, this was what was causing the failure. I found this thread: https://bitbucket.org/hvisser/android-apt/issues/22/dagger-robolectric-gradle-apt-source-not

PR #750, opened 2016-10-24
Removes LostClientManager connection state

ecgreb:
In an effort to simplify location client management `LostClientManager` no longer actively tracks its connection state. Instead it relies on `LostApiClient.isConnected()` to return proper connection state.

`LostClientManager.getClient()` now always returns the client instance and is no longer nullable. Null checks on the client have been converted to calls to `isConnected()`.


ecgreb:
I know I took a shortcut on some of the tests here and not all code paths are covered in the presenter regarding location initialization but we can continue to improve coverage as we iterate on location management in the app now that its out of the activity

PR #752, opened 2016-10-24
Dont invoke client connect when already connected

### LINKED ISSUE #751: Crash when mock routing enabled

sarahlensing:
This change fixes a crash caused when mock routing was enabled. Previously, calls to connect an already connected client created an infinite loop in `MainActivity#checkPermissionAndEnableLocation`. `LostApiClient#connect` invoked `ConnectionCallbacks#onConnected`. This would trigger queued runnables to run in `LostClientManager` which would start the loop again...now the underlying client will not be called if it is already connected.

Closes #751

### DIFF
--- a/app/src/main/kotlin/com/mapzen/erasermap/model/LostClientManager.kt
+++ b/app/src/main/kotlin/com/mapzen/erasermap/model/LostClientManager.kt
@@ -44,7 +44,7 @@ class LostClientManager(val application: Application, locationFactory: LocationF
   }
 
   override fun connect() {
-    if (state == State.CONNECTING) {
+    if (state == State.CONNECTING || state == State.CONNECTED) {
       return
     }
     state = State.CONNECTING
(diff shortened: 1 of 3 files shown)


app/src/main/kotlin/com/mapzen/erasermap/model/LostClientManager.kt
```diff
@@ -44,7 +44,7 @@ class LostClientManager(val application: Application, locationFactory: LocationF
   }
 
   override fun connect() {
-    if (state == State.CONNECTING) {
+    if (state == State.CONNECTING || state == State.CONNECTED) {
```
ecgreb: This is going to conflict with https://github.com/mapzen/eraser-map/pull/750 where I removed tracking the connection state locally in favor of calling `lostClient.isConnected()`. Would be be able check the state of the client here instead?

sarahlensing: Looks like this change is not needed with https://github.com/mapzen/eraser-map/pull/750 merged

PR #760, opened 2016-10-27
Search result restoration

### LINKED ISSUE #759: Search Results List Not Restored On Orientation
* **App build number:**
0.13.0-SNAPSHOT

Rotating device when search results list shown would redisplay search results list

* **What happened instead?**
List was hidden

* **Steps to reproduce:**
Click search results list button and rotate device. Results will not be shown after orientation change.

sarahlensing:
- Updates how we handle restoring the options menu so that we don't try to call methods on it before it has been created
- Fixes bug which prevented options menu buttons from being restored on device rotation when viewing search results
- Fixes bug which didnt restore search results list view on device orientation
- Fixes bug which caused the settings button to show (upon back pressing to search results) when rotating the device while viewing the route preview, route preview directions, route, and route directions screens

Closes #759

### DIFF
--- a/app/src/main/kotlin/com/mapzen/erasermap/presenter/MainPresenterImpl.kt
+++ b/app/src/main/kotlin/com/mapzen/erasermap/presenter/MainPresenterImpl.kt
@@ -192,6 +192,10 @@ open class MainPresenterImpl(val mapzenLocation: MapzenLocation, val bus: Bus,
     mainViewController?.hideSettingsBtn()
     mainViewController?.setOptionsMenuIconToList()
     updateViewAllAction(searchResults)
+    if (resultListVisible && searchResults?.features != null) {
+      val features = searchResults?.features as List<Feature>
+      mainViewController?.onShowAllSearchResultsList(features)
+    }
   }
 
   override fun onRestoreMapState() {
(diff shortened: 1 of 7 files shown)


app/src/main/kotlin/com/mapzen/erasermap/presenter/MainPresenterImpl.kt
```diff
@@ -156,7 +160,7 @@ open class MainPresenterImpl(val mapzenLocation: MapzenLocation, val bus: Bus,
   }
 
   private fun onRestoreViewStateSearchResults() {
-    mainViewController?.hideSettingsBtn()
+    // Do nothing.
```
msmollin: Why override if we're doing nothing?

msmollin: Per our discussion in slack, this is because we need the `when` statement to be exhaustive. Let's try out moving these into just anonymous blocks that get executed at the actual `when` statement rather than separate functions. Should make the code more concise and readable.

sarahlensing: Updated.

PR #767, opened 2016-10-31
Dont show action bar on rotation in route preview mode

### LINKED ISSUE #765: Tapping on search bar after rotation while routing preview is up leaves route line artifacts
- **Device name:** 
  Nexus 5
- **Android Version:** 
  4.4.4
- **App build number:**
  master-1795
- **What did you expected to happen?** 
  The search bar not to be available just rotating the device while in routing preview
- **What happened instead?**
  I can tap on the search screen and search for a new place. This dumps me out of the routing preview flow and seems to orphan drawn poly lines on the map, and leaves a weird gray pin as well over your current location.
- **Steps to reproduce:**
- Route to a destination. 
- Rotate the device.
- Notice the search bar is available for a few seconds, and tap into it to search.
- **Attach a screenshot**
  It's hard to grab a screen shot mid test, but here's the ultimate outcome of the above process.
  screenshot_2016-10-28-16-30-31

sarahlensing:
This PR prevents the action bar (search bar) from showing when the device is rotated while in route preview mode by hiding it as soon as the action bar is created (previously it wasnt hidden until the route was successfully fetched from the server). It also extracts out some logic for handling what happens when a route is successfully fetched and moves it into the `MainPresenter`/adds test coverage.

Closes #765

### DIFF
--- a/app/src/main/kotlin/com/mapzen/erasermap/presenter/MainPresenterImpl.kt
+++ b/app/src/main/kotlin/com/mapzen/erasermap/presenter/MainPresenterImpl.kt
@@ -471,10 +471,8 @@ open class MainPresenterImpl(val mapzenLocation: MapzenLocation, val bus: Bus,
   }
 
   override fun success(route: Route) {
-    mainViewController?.hideProgress()
-    routeManager.route = route
+    handleRouteRetrieved(route)
     generateRoutingMode(false)
-    mainViewController?.drawRoute(route)
     waitingForRoute = false
   }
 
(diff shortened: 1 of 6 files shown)


app/src/main/kotlin/com/mapzen/erasermap/presenter/MainPresenterImpl.kt
```diff
@@ -586,6 +592,18 @@ open class MainPresenterImpl(val mapzenLocation: MapzenLocation, val bus: Bus,
     routeManager.fetchRoute(callback)
   }
 
+  override fun onRouteSuccess(route: Route) {
```
ecgreb: `MainPresenterImpl` also implements `RouteCallback` itself since reroute logic is handled in this class.

https://github.com/mapzen/eraser-map/blob/master/app/src/main/kotlin/com/mapzen/erasermap/presenter/MainPresenterImpl.kt#L467

Seems like there may be an opportunity to combine this method and `success(Route)` or maybe at least have one invoke the other. This could possibly avoid some code dupe and/or divergent behavior for route vs. reroute?

PR #771, opened 2016-11-02
Improves route preview landscape layout

### LINKED ISSUE #764: Reorient Camera button is always visible on routing preview which causes layout issues in horizontal orientation on smaller screen devices
- **Device name:** 
  Nexus 5
- **Android Version:** 
  4.4.4
- **App build number:**
  master-1795
- **Steps to Reproduce:**
- Route to a location.
- Rotate the device to horizontal orientation. 

There's a number of issues here which can be seen in the attached screen shot. 1 is the camera reorientation button is constantly visible even though I haven't reoriented the camera. This cause the "find me" button to get hidden and causes UI weirdness.

Second is the layout for routing preview on smaller screen devices is not terribly functional due to how large the font and iconography are.

screenshot_2016-10-28-16-28-56

ecgreb:
Improves route preview layout in landscape mode.

image

* Introduces modified route preview layout in landscape mode with smaller text and margins to show more of the map.
* Moves distance and time onto either side of route mode widget in landscape mode.
* Does not show "Try a different mode?" text in landscape mode when no route is found.
* Adds null checks for layout components that only exist in portrait or landscape.
* Breaks down route preview layout code into smaller re-usable chunks.
* Extracts more values into `dimens.xml` to vary portrait vs. landscape.
* Fixes issue where compass and find me button were overlapping in landscape mode.

Closes #764


souperneon:
Looks good @sarahlensing @ecgreb My only comment would be, separating distance and time makes it harder for the user - they have to look left and right before picking a mode. Don't have a more elegant solution off the top of my head that would be an easy fix. Wondering if others feel he same?

ecgreb:
@souperneon I agree but also couldn't think of a better option that still conserves vertical space. Maybe lets go with this for now and revisit in the next release?

PR #777, opened 2016-11-08
Improve test coverage in MainPresenter

sarahlensing:
This PR moves logic for exiting navigation, hiding route preview, and centering on features out of the `MainActivity` and into the `MainPresenter` as well as increases test coverage. It replaces large methods with simple calls to UI elements in `MainActivity`. It also adds a fix to reset the tilt when navigation mode is exited. There is a lot more work to do here, this is just a start...

### DIFF
--- a/app/src/main/kotlin/com/mapzen/erasermap/controller/MainActivity.kt
+++ b/app/src/main/kotlin/com/mapzen/erasermap/controller/MainActivity.kt
@@ -172,7 +172,7 @@ class MainActivity : AppCompatActivity(), MainViewController,
 
     override fun onNewIntent(intent: Intent?) {
         if (intent?.getBooleanExtra(NotificationCreator.EXIT_NAVIGATION, false) as Boolean) {
-            exitNavigation()
+            presenter.onExitNavigation()
             if((intent?.getBooleanExtra(NotificationBroadcastReceiver.VISIBILITY, false)
                     as Boolean).not()) {
                 moveTaskToBack(true)
@@ -580,7 +580,7 @@ class MainActivity : AppCompatActivity(), MainViewController,
         return layoutParams
     }
 
-    private fun layoutFindMeAlignBottom() {
+    override fun layoutFindMeAlignBottom() {
         val layoutParams = baseFindMeParams()
         layoutParams.addRule(RelativeLayout.ALIGN_PARENT_BOTTOM)
         val margin = resources.getDimensionPixelSize(R.dimen.em_find_me_button_margin_bottom)
@@ -952,14 +952,22 @@ class MainActivity : AppCompatActivity(), MainViewController,
 
     override fun hideRoutePreview() {
         if((findViewById(R.id.route_mode) as RouteModeView).visibility != View.VISIBLE) {
-            supportActionBar?.show()
+            showActionBar()
             routeManager.reverse = false
-            findViewById(R.id.route_preview)?.visibility = View.GONE
+            hideRoutePreviewView()
             hideRoutePins()
             val features = arrayListOf(presenter.currentFeature) as List<Feature>
             layoutAttributionAboveSearchResults(features)
             layoutFindMeAboveSearchResults(features)
         }
+    }
+
+    override fun hideRoutePreviewView() {
+        routePreviewView.visibility = View.GONE
+    }
+
+    override fun showActionBar() {
+        supportActionBar?.show()
     }
 
     override fun route() {
@@ -1091,7 +1099,7 @@ class MainActivity : AppCompatActivity(), MainViewController,
         }
     }
 
-    private fun setDefaultCamera() {
+    override fun setDefaultCamera() {
         mapzenMap?.cameraType = CameraType.ISOMETRIC
     }
 
@@ -1108,12 +1116,20 @@ class MainActivity : AppCompatActivity(), MainViewController,
         presenter.routingEnabled = false
         setDefaultCamera()
         checkPermissionAndEnableLocation()
-        routeModeView.visibility = View.GONE
+        hideRouteModeView()
         supportActionBar?.hide()
         routeModeView.route = null
-        routeModeView.hideRouteIcon()
+        hideRouteIcon()
         routeModeView.hideResumeButton()
         hideReverseGeolocateResult()
+    }
+
+    override fun hideRouteIcon() {
+        routeModeView.hideRouteIcon()
+    }
+
+    override fun hideRouteModeView() {
+        routeModeView.visibility = View.GONE
     }
 
     override fun overridePlaceFeature(feature: Feature) {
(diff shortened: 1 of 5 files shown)


ecgreb:
Enabling the debug menu by searching for "!!!!!!!!" while testing this branch led to a crash documented here https://github.com/mapzen/eraser-map/issues/780.

sarahlensing:
Pushed a fix for this crash

app/src/main/kotlin/com/mapzen/erasermap/controller/MainActivity.kt
```diff
@@ -711,11 +711,15 @@ class MainActivity : AppCompatActivity(), MainViewController,
     }
 
     override fun addSearchResultsToMap(features: List<Feature>?, activeIndex: Int) {
-        if (features == null) {
+        if (features == null || features.size == 0) {
```
ecgreb: In the spirit of this refactor what do you think about moving both of these checks into the presenter so they can also be covered with unit tests? If `features` is either null or size 0 this method would never get called. Then the `features` param could also be declared non-null.

Also it seems like there is an opportunity to clean up the relationship between this method `MainActivity.addSearchResultsToMap(features: List<Feature>?, activeIndex: Int)` and `MainActivity.showSearchResults(features: List<Feature>?)`.

Both get called from the `MainPresenter` at different times. In addition`MainActivity.showSearchResults(features: List<Feature>?)` directly invokes `MainActivity.addSearchResultsToMap(features: List<Feature>?, activeIndex: Int)`.

Some questions that come to mind:
* What is the difference between these two methods?
* Why is one called over the other in `MainPresenter`?
* Why does one call the other rather than the presenter calling both in sequence?
* Can they be combined into a single method?
* Can they be renamed/documented so the purpose is more clear?

We don't have to solve this issue right now but I am interested to get your thoughts. It could be the subject of a good follow-up pull request to this one.

app/src/test/kotlin/com/mapzen/erasermap/presenter/MainPresenterTest.kt
```diff
@@ -536,6 +607,48 @@ class MainPresenterTest {
         assertThat(mainController.routeRequestCanceled).isTrue()
     }
 
+    @Test fun onBackPressed_shouldShowActionBarInStateRoutePreview() {
+        vsm.viewState = ROUTE_PREVIEW
+        mainController.isActionBarHidden = false
```
ecgreb: Should set `mainController.isActionBarHidden = true` here to ensure flag is toggled correctly when back is pressed.

app/src/main/kotlin/com/mapzen/erasermap/presenter/MainPresenterImpl.kt
```diff
@@ -688,4 +705,27 @@ open class MainPresenterImpl(val mapzenLocation: MapzenLocation, val bus: Bus,
     mainViewController?.drawRoute(route)
     mainViewController?.hideProgress()
   }
+
+  private fun featuresExist(features: List<Feature>?): Boolean {
+    return features != null || features!!.size > 0
```
ecgreb: Should this be `&&` operator?

Currently the second half of conditional will never evaluate unless features is null in which case the `!!` operator will throw a NPE.

app/src/main/kotlin/com/mapzen/erasermap/controller/MainActivity.kt
sarahlensing: Moving these checks into the presenter is on my radar/"next to refactor" list. I didn't do it here because I didn't want to add more changes to this PR but I completely agree that they should be moved into the presenter and covered with tests.

The reason we have these two methods is because we need one method to adjust the views on the screen when search results should be shown (make the search results view visible, adjust the attribution & find me btn)...thats `MainActivity.showSearchResults(features: List<Feature>?)`. And then we have a more specific method which handles updating the map's search result pins/position/zoom (`MainActivity.addSearchResultsToMap(features: List<Feature>?, activeIndex: Int)`) when the search results pager is adjusted.

When search results are first shown, or when we are backing out of route preview, we call `showSearchResults` so that hidden/moved views will be properly adjusted and search results on the map will be shown. But, when the user pages the search results, we don't need to update all the other views on the screen, only remove/re-add the search result pins so that the newly select result is "active"

I think that the logic in `showSearchResults` should  be moved to the presenter so that it can handle calling both and we can add test coverage for this.

I think there wouldn't be any harm in combining them into a single method but I would lean towards keeping them separate and documenting them. This way, we aren't making unnecessary calls to update ancillary views when only the active search index changes. I think `addSearchResultsToMap` is clear but maybe `showSearchResults` could  become `prepareAndShowSearchResultsViews`.

I will be addressing this in the next batch of EM refactoring.

app/src/test/kotlin/com/mapzen/erasermap/presenter/MainPresenterTest.kt
sarahlensing: Fixed

app/src/main/kotlin/com/mapzen/erasermap/presenter/MainPresenterImpl.kt
sarahlensing: Eek, yep it should be. Fixed.

app/src/test/kotlin/com/mapzen/erasermap/presenter/MainPresenterTest.kt
```diff
@@ -466,19 +468,88 @@ class MainPresenterTest {
...
+        features.add(Feature())
+        features.add(Feature())
+        val feature = SimpleFeature.create("1", "1", "", "",
+            "", "", "", "", "", "", "", 0.0, "", "", 40.0, 70.0)
+        features.add(feature.toFeature())
+        result.features = features
+        mainController.currentSearchItemPosition = 2
+        presenter.onSearchResultsAvailable(result)
+        presenter.onSearchResultSelected(0)
+        assertThat(mainController.currentSearchItemPosition).isEqualTo(2)
```
ecgreb: Search result selection seems to work fine in the app but I'm not sure I follow the logic in the test.

If search result 0 is selected, why would mainController.currentSearchItemPosition be equal to 2? Shouldn't it get set to 0 on selection?

app/src/test/kotlin/com/mapzen/erasermap/presenter/MainPresenterTest.kt
```diff
@@ -466,19 +468,88 @@ class MainPresenterTest {
...
+        val feature = SimpleFeature.create("1", "1", "", "",
+            "", "", "", "", "", "", "", 0.0, "", "", 40.0, 70.0)
+        features.add(feature.toFeature())
         result.features = features
+        mainController.currentSearchItemPosition = 2
         presenter.onSearchResultsAvailable(result)
         presenter.onSearchResultSelected(0)
-        assertThat(mainController.isCenteredOnCurrentFeature).isTrue()
+        assertThat(mainController.lngLat?.longitude).isEqualTo(70.0)
+        assertThat(mainController.lngLat?.latitude).isEqualTo(40.0)
```
ecgreb: Same question here. If search result 0 is selected why are we zooming the map to the lat/lng for the result at index 2?

app/src/test/kotlin/com/mapzen/erasermap/presenter/MainPresenterTest.kt
sarahlensing: This is a little confusing because there isn't a scenario in the app where the current search item position would be 2 but the search result selected would be 0...it would also be 2. I will update `presenter.onSearchResultSelected(0)` to select the second result. 

The current position doesn't get set to 0 because the index here is only used to determine which search pin should be active.

PR #783, opened 2016-11-10
Preserve search pager position on rotation & backing out of route preview

### LINKED ISSUE #778: Search Result Pager Position Not Preserved
* **App build number:**
0.13.0-SNAPSHOT

When I scroll the search result pager past the first position and then go into route preview, I expect that backing out of route preview would keep me at the search result position I scrolled to.

* **What happened instead?**
The search result pager was put back to the first position.

* **Steps to reproduce:**
Search for something. Scroll the pager past the first position. Go into route preview. Back out of route preview.

sarahlensing:
This PR stores the search pager position in `MainPresenter` so that it can be used to restore the search results view when either the device is rotated or the user backs out of route preview mode. It also adds tests for this case.

I chose to leave the logic in `MainActivity` to keep the changes in this PR minimal and plan to move this code into the presenter in a separate PR as part of the ongoing test refactoring.

Closes #778

### DIFF
--- a/app/src/main/kotlin/com/mapzen/erasermap/presenter/MainPresenterImpl.kt
+++ b/app/src/main/kotlin/com/mapzen/erasermap/presenter/MainPresenterImpl.kt
@@ -68,6 +68,7 @@ open class MainPresenterImpl(val mapzenLocation: MapzenLocation, val bus: Bus,
     set(value) {
       confidenceHandler.reverseGeoLngLat = value
     }
+  override var currentSearchIndex: Int = 0
 
   private var searchResults: Result? = null
   private var destination: Feature? = null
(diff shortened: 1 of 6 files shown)


app/src/main/kotlin/com/mapzen/erasermap/presenter/MainPresenterImpl.kt
```diff
@@ -64,6 +64,7 @@ open class MainPresenterImpl(val mapzenLocation: MapzenLocation, val bus: Bus,
   override var resultListVisible = false
   override var reverseGeo = false
   override var reverseGeoLngLat: LngLat? = null
+  override var currentSearchIndex: Int = 0
```
ecgreb: This flag should be reset to 0 when the current set of search results is cleared.

Otherwise the next time a search is executed the current index from the previous set of results is displayed by default.

ecgreb:
Changes look good just needs a rebase.

PR #789, opened 2016-11-15
Restore map position and zoom for reverse geo back press

### LINKED ISSUE #779: Center Map On Reverse Geo Feature
* **App build number:**
0.13.0-SNAPSHOT

After long pressing on the map to reverse geolocate a feature, I enter route preview mode. I would expect that backing out of route preview would center the map on this selected feature like it does for search results.

* **What happened instead?**
The map was zoomed out and not centered on the selected feature.

* **Steps to reproduce:**
Long press on map to produce pin on map. Enter route preview mode. Back out of route preview mode.

sarahlensing:
### Overview
Caches the map position and zoom when route preview is generated. When back is pressed for reverse geo requests, the presenter uses these values to restore the map to the previous position and zoom.

### Proposed Changes
Adds two new methods to `MainViewController` for retrieving the map zoom and position.

Closes #779

### DIFF
--- a/app/src/main/kotlin/com/mapzen/erasermap/presenter/MainPresenterImpl.kt
+++ b/app/src/main/kotlin/com/mapzen/erasermap/presenter/MainPresenterImpl.kt
@@ -71,6 +71,8 @@ open class MainPresenterImpl(val mapzenLocation: MapzenLocation, val bus: Bus,
       confidenceHandler.reverseGeoLngLat = value
     }
   override var currentSearchIndex: Int = 0
+  override var mapPosition: LngLat? = null
+  override var mapZoom: Float? = null
 
   private var searchResults: Result? = null
   private var destination: Feature? = null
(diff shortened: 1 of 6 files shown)


app/src/main/kotlin/com/mapzen/erasermap/presenter/MainPresenterImpl.kt
```diff
@@ -391,6 +395,8 @@ open class MainPresenterImpl(val mapzenLocation: MapzenLocation, val bus: Bus,
     if (searchResults != null) {
       if (reverseGeo) {
         mainViewController?.showReverseGeocodeFeature(searchResults?.features)
+        mainViewController?.setMapPosition(mapPosition as LngLat, 0)
```
ecgreb: This could lead to a class cast exception if `mapPosition` is `null`

app/src/main/kotlin/com/mapzen/erasermap/presenter/MainPresenterImpl.kt
```diff
@@ -391,6 +395,8 @@ open class MainPresenterImpl(val mapzenLocation: MapzenLocation, val bus: Bus,
...
+        mainViewController?.setMapZoom(mapZoom as Float)
```
ecgreb: This could lead to a class cast exception if `mapZoom` is `null`

PR #791, opened 2016-11-22
Improve test coverage

sarahlensing:
### Overview
This PR moves pure java code out of `MainActivity` and into `MainPresenterImpl` as well as adds more test coverage for the moved logic.

### Proposed Changes
Map listener handling (for `onMapPressed` and `onMapDoubleTapped`) has been moved into the presenter. In addition, `showSearchResults` and `addSearchResultsToMap` have both been moved into the presenter and replaced by smaller methods in the `MainViewController`.


app/src/main/kotlin/com/mapzen/erasermap/controller/MainActivity.kt
```diff
@@ -652,6 +622,9 @@ class MainActivity : AppCompatActivity(), MainViewController,
      * to show or hide debug settings in settings fragment
      */
     override fun toggleShowDebugSettings() {
+        if (!AndroidAppSettings.SHOW_DEBUG_SETTINGS_QUERY.equals(searchController.searchView?.query.toString())) {
```
ecgreb: In a recent pull request this logic was moved into the presenter with the idea being that `toggleShowDebugSettings()` should not be called unless this check had already been made. Is that not the case anymore?

https://github.com/mapzen/eraser-map/pull/787

sarahlensing: Erg, that is still the case...what happened is that I refactored the kotlin syntax and then when resolving the rebase conflicts not realized that it had been moved. 

Removed.

app/src/main/kotlin/com/mapzen/erasermap/presenter/MainPresenterImpl.kt
```diff
@@ -412,6 +487,53 @@ open class MainPresenterImpl(val mapzenLocation: MapzenLocation, val bus: Bus,
...
+    if (features == null) {
+      return
+    }
+
+    mainViewController?.hideReverseGeolocateResult()
+    mainViewController?.showSearchResultsView(features)
+    addSearchResultsToMap(features, activeIndex)
+    mainViewController?.layoutAttributionAboveSearchResults(features)
+    mainViewController?.layoutFindMeAboveSearchResults(features)
+    mainViewController?.toggleShowDebugSettings()
```
ecgreb: I think this call to `toggleShowDebugSettings()` should be removed. Its already invoked from `MainPresenter.onQuerySubmit(...)` with the appropriate check for the matching term. Calling it again from here causes the debug menu to toggle for any search query or when any autocomplete term is selected from the drop down list.

sarahlensing: You're right, it can be removed. It wasn't causing unexpected results because this method was never reached for the debug query term.

I've removed this call.

PR #825, opened 2017-06-30
Fix Exit Navigation Crash

### LINKED ISSUE #824: Crash when exiting navigation while app is in background
* **Device name:** Galaxy S8 & Nexus 5
* **Android Version:** 7.0 & 4.4 respectively
* **App build number:** 0.13.1
When exiting background routing, it should just exit and not crash
* **What happened instead?**
Exiting routing from the pinned notification will cause a dialog to eventually (a few seconds) appear stating eraser map has crashed
* **Steps to reproduce:**
Start voice navigation along a route, background the app, and then exit the navigation using the notification. It doesn't matter what route is currently happening.

sarahlensing:
### Overview
- Optimizes how the app handles enabling the location layer. Now it is not enabled when exiting routing mode from the notification if the app is in the background
- Updates the app to unregister the routing location request when routing mode exited from the notification
- Also stops the speaker when exiting routing
- Centers map on current location with default zoom when routing exited
- Updates map overlay mode to "classic" (so that find me button is always at bottom right instead of just when backing out of routing mode)

Underlying Lost issue has been resolve and will be available in a future version of the SDK https://github.com/mapzen/lost/pull/227

Closes #824

### DIFF
--- a/app/src/main/kotlin/com/mapzen/erasermap/presenter/MainPresenter.kt
+++ b/app/src/main/kotlin/com/mapzen/erasermap/presenter/MainPresenter.kt
@@ -32,6 +32,7 @@ interface MainPresenter {
     var poiTapPoint: FloatArray?
     var poiTapName: String?
     var poiCoordinates: LngLat?
+    var willPauseImmediately: Boolean?
 
     fun onSearchResultsAvailable(result: Result?)
     fun onReverseGeocodeResultsAvailable(searchResults: Result?)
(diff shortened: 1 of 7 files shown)


app/src/main/kotlin/com/mapzen/erasermap/presenter/MainPresenter.kt
```diff
@@ -32,6 +32,7 @@ interface MainPresenter {
     var poiTapPoint: FloatArray?
     var poiTapName: String?
     var poiCoordinates: LngLat?
+    var willPauseImmediately: Boolean?
```
tallytalwar: Maybe a better name can be used here to clarify the context its used, not sure what though!

